PR #2, opened 2016-05-19
Add minimal_lstm files and README.md

danabo:
melody_rnn directory is created. This will be the project directory for recurrent melody generating models. To start it off, a minimal LSTM (long short-term memory) model is added.

This code shows how to build a recurrent graph in TensorFlow using either dynamic loops (dynamic_rnn) or static unrolling in chunks (state_saving_rnn).

### DIFF
--- a/melody_rnn/README.md
+++ b/melody_rnn/README.md
@@ -1,0 +1,11 @@
+# Melody RNN Models
+
+This directory contains a number of RNN (recurrent neural network) models that can generate melodies. These models fall under the category of language models, in that they output the probability of the next note given a sequence of previous notes in a modely. Melodies can be sampled from these models by searching for the most likely sequence of notes (using beam search for example).
+
+## Minimal RNN
+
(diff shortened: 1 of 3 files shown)


fredbertsch:
Question: Does this fit better in tensorflow/models instead of tensorflow/magenta? (I think I like it here better, but I want a record of the conversation about it.)

2nd question: Why is it important that the models be RNN models? Can't we call it melody_models instead?

danabo:
1) I thought we were putting everything under Magenta team's directory. Ask
Doug about this.

2) I was copying the structure of our internal code, which seems to be
divided up by project and model type. But yes I agree lets organize by
problem space.

fredbertsch:
No, not everything is going under Magenta's GitHub. An easy counterexample are new ops. Those need to go under tensorflow. Tensorflow already has a home for models. We need to decide if the costs of having two homes for TF models is worth it.

fredbertsch:
Discussed with the rest of the team. Let's put this under https://github.com/tensorflow/models instead. They're happy to have quirky, specialized models there, and it'd be nice to have models in one place.

danabo:
Okay I will close this pull request then.

PR #4, opened 2016-05-26
Add music protobuf and BUILD file.

adarob:


fredbertsch:
Add the license text to the build and proto files. (This is tricky on my phone, so I'll look closer tomorrow.)

magenta/protobuf/music.proto
```diff
@@ -0,0 +1,125 @@
...
+
+package tensorflow.magenta;
+
+// A message containing a symbolic music sequence. The design is largely
+// based on MIDI but it should be able to represent any music sequence.
+// For details see https://www.midi.org/specifications.
+message NoteSequence {
+  // NOTE: The first three fields are duplicated from SequenceFile to allow
+  // creation of portable SSTables of Sequence protos.
+  // Unique id. See data_utils.h GenerateSequenceId().
```
fredbertsch: Last sentence isn't help.

magenta/protobuf/music.proto
```diff
@@ -0,0 +1,125 @@
...
+  string id = 1;
+  // The path of the file relative to the root of the collection.
+  string filename = 2;
+  // The collection from which the file comes. This can be shorthand e.g.
+  // "bach". One purpose is to allow for easy selection of all or some files
+  // from a particular source.
+  string collection = 3;
```
fredbertsch: Presumably this is not a globally unique id. Unless that's the case, let's call it "collection_name" to make it sound less like a centrally managed name.

magenta/protobuf/music.proto
```diff
@@ -0,0 +1,125 @@
...
+
+  // MIDI ticks per beat, also known as resolution or PPQ ("parts per quarter").
+  // There is no widely-used default. 220 is assumed per the choice made in
+  // third_party/py/pretty_midi.
```
fredbertsch: The last sentence needs to be rewritten. "If unspecified, 220 is assumed."

Not that this is the right place to complain, but 220 is a weird default value. Why not make triplets work perfectly? I'd vote for 60.

magenta/protobuf/music.proto
```diff
@@ -0,0 +1,125 @@
...
+      B = 11;
+    }
+
+    enum Mode {
+      MAJOR = 0;
+      MINOR = 1;
+    }
+  }
+
+  message TempoChange {
```
fredbertsch: The first one isn't a TempoChange, it's a Tempo. Let's call them all Tempos.

magenta/protobuf/music.proto
```diff
@@ -0,0 +1,125 @@
...
+    // Time in seconds.
+    double time = 1;
+    // Tempo in beats per minute.
+    double bpm = 2;
+  }
+
+  // Stores MIDI PitchBend data. See the MIDI specfication for details.
```
fredbertsch: Can you link to the portion of the MIDI spec? There's a bunch of oddness in this message that'd be useful to look up.

magenta/protobuf/music.proto
lasconic: FWI, [MuseScore](https://musescore.org) uses 480 and quintuplets work perfectly too.

magenta/protobuf/music.proto
adarob: Removed this and other internal references.

magenta/protobuf/music.proto
adarob: Are you referring to the id? It is intended to be globally unique.

magenta/protobuf/music.proto
adarob: @douglaseck thoughts?

magenta/protobuf/music.proto
adarob: Fixed.

magenta/protobuf/music.proto
adarob: Unfortunately the spec is in a non-linkable format.

magenta/protobuf/music.proto
fredbertsch: I was talking about "collection". Presumably, there are a finite number of interesting MIDI collections out there, but, once we're in OSS land, there's no central naming authority for them. So, I don't think they will be unique. (Unless I'm missing something?) You don't say anything about this being either unique or human readable. A comment about that would be helpful. If it's not unique, then a name that emphasizes the field's purpose would also be helpful.

magenta/protobuf/music.proto
fredbertsch: Looks like @douglaseck is okay with this. (Offline discussion)

PR #7, opened 2016-06-02
Fix artist name

sroecker:


sroecker:
I signed it!

fredbertsch:
Thanks for the fix!

PR #13, opened 2016-06-05
Update bazel-bin filepath in 'Sample usage' section

slamphear:
After building the convert_midi_dir_to_note_sequences script, the resulting files are directly in bazel-bin/magenta (and not in bazel-bin/magenta/scripts).


adarob:
Thanks!

PR #35, opened 2016-06-10
Fix grammar for basic_rnn/README

ZuzooVn:


danabo:
Thanks for the edits.

PR #38, opened 2016-06-10
Using chained comparison operators

ZuzooVn:
Unlike most languages, Python supports chained comparison operators and
it evaluates them as they would be evaluated in normal mathematics.


danabo:
Looks good to me.

PR #39, opened 2016-06-10
Fix typo

ZuzooVn:


danabo:
Not a typo. The preceding 'r' makes it a raw string, so that back slashes do not need to be escaped.

PR #54, opened 2016-06-14
Each model has its own script for building datasets.

danabo:
convert_sequences_to_melodies.py is turned into a library. basic_rnn has
create_dataset.py which imports sequence_to_melodies and builds the
dataset. Models will define their own encoders in their
create_dataset.py scripts. In this case, a default encoder is defined in
sequence_to_melodies.py and used by basic_rnn.


magenta/lib/sequence_to_melodies.py
```diff
@@ -11,16 +11,21 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-"""Encoder functions for converting a melody list to SequenceExample.
+"""Converts Sequence protos to SequenceExample protos.
```
adarob: NoteSequence

magenta/lib/sequence_to_melodies.py
```diff
@@ -51,7 +56,7 @@ def basic_one_hot_encoder(melody, steps_per_beat=4, min_note=48, max_note=84,
 
   Returns:
     sequence_example: A SequenceExample proto containing inputs and labels sequences.
-    encoder_information: Information needed to decoded encoder output.
+    reconstruction_data: Information needed to decoded encoder output.
```
adarob: s/decoded/decode/g

magenta/lib/sequence_to_melodies.py
```diff
@@ -90,4 +95,57 @@ def basic_one_hot_encoder(melody, steps_per_beat=4, min_note=48, max_note=84,
...
+        protos.
+    train_output: String path to TFRecord file that training samples will be
+        saved to.
+    eval_output: If set, string path to TFRecord file that evaluation samples
+        will be saved to. Omit this argument to not produce an eval set.
+    eval_ratio: Fraction of input that will be saved to eval set. A random
+        partition is chosen, so the actual train/eval ratio will vary.
+  """
+
+  reader = tf.python_io.tf_record_iterator(sequences_file)
```
adarob: use note_sequence_record_iterator

i'm going to push this into TF as a general wrapper

danabo: That's fine. Just wondering though, why have proto-specific wrappers when the TF API supports reading and writing arbitrary serialized protos (and arbitrary bytes)?

magenta/lib/sequence_to_melodies.py
danabo: Done.

magenta/lib/sequence_to_melodies.py
danabo: Done.

magenta/lib/sequence_to_melodies.py
danabo: We discussed offline. note_sequence_record_iterator will eventually turn into a wrapper that takes arbitrary protos instead of arbitrary bytes.

adarob:
LGTM

PR #60, opened 2016-06-20
Link to whole basic_rnn folder to show the readme plus the files

devinrhode2:
When reading the readme for basic_rnn I had to stop and see where convert_sequences_to_melodies.py is, but if we instead link to the folder and not directly to it's readme, then those files will be visible right at the top. Provides good context for the readme, mirrors how all github repos are layed out.


fredbertsch:
I like that, too. Thanks!

PR #70, opened 2016-06-29
Add a default of 220 ticks_per_beat to avoid broke MIDI files.

douglaseck:


fredbertsch:
proto3 doesn't have defaults

PR #64, opened 2016-06-24
rename nonexisting filename. clarify explanation.

vug:
convert_sequences_to_melodies.py -> basic_rnn_create_dataset.py

PR #76, opened 2016-06-30
Rename nonexisting filename. (2)

vug:
and clarify the explanation a little bit. This is a clone of the previous PR #64 with correct CLA.


fredbertsch:
Thanks for fixing this (twice). :)

PR #80, opened 2016-07-04
import StringIO in both Python 2 & 3

vug:
Python 3 removed StringIO and cStringIO modules. Instead there is
io module. From https://docs.python.org/3.0/whatsnew/3.0.html
"The StringIO and cStringIO modules are gone. Instead,
import the io module and use io.StringIO or io.BytesIO for text
and data respectively."

### DIFF
--- a/magenta/lib/midi_io.py
+++ b/magenta/lib/midi_io.py
@@ -17,8 +17,12 @@ Input and output wrappers for converting between MIDI and other formats.
 """
 
 from collections import defaultdict
-from cStringIO import StringIO
 import sys
+if sys.version_info.major <= 2:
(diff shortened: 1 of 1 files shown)


vug:
Hi! I didn't see any info on project README about which Python version is supported. I'm using Python 3.5 and encountered ImportError while importing StringIO and fixed it with this commit.

fredbertsch:
Hello! Thanks for waiting. I was out of town on holiday...

This looks great. We're running Python 2.7.6 internally, but we definitely want to support Python 3 as well. We're working on setting up Python 3 tests, for example. 

Meanwhile, this is a great step in the right direction. LGTM.

PR #84, opened 2016-07-06
Disable basic_rnn_train_test.

cghawthorne:
Should turn back on when the test doesn't time out.


fredbertsch:
LGTM

danabo:
I'll have to make the test more lenient. Converting over to TFLearn should fix this (if I understand correctly that it automates boilerplate model testing).

PR #85, opened 2016-07-06
fix linter warning

cghawthorne:


magenta/lib/midi_io.py
```diff
@@ -18,6 +18,7 @@
 
 from collections import defaultdict
 import sys
+# pylint: disable=g-import-not-at-top
```
fredbertsch: Please re-enable the pylint warning after this import is complete.

cghawthorne: Done

adarob:
LGTM

PR #86, opened 2016-07-08
PixelRNN review and assets

kastnerkyle:
A review of [Pixel Recurrent Neural Networks](https://arxiv.org/abs/1601.06759)


kastnerkyle:
```
I signed it!
```

PR #92, opened 2016-07-12
Tweaked license based on Ishaan's license

kastnerkyle:
Changed license to match Ishaan's [see LICENSE here](https://github.com/igul222/pixel_rnn)


fredbertsch:
Unfortunately, Google is requiring us to use a unform license in this repo. We're looking into finding an alternative.

kastnerkyle:
Ok - I think I can relicense to apache 2.0 . Can we confirm that is
possible?

If not we can just point to Ishaan's code and pull this file.

PR #93, opened 2016-07-12
Removing bad license

fredbertsch:
Unfortunately, we're not able to support BSD licensed files in this repo. Removing this while we figure out an alternative.


cghawthorne:
LGTM

PR #94, opened 2016-07-13
Fix url for Conditional Pixel CNN

carpedm20:
"more recent paper" is directing the original paper not the recent one.


cinjon:
Thanks carpedm20.

PR #103, opened 2016-07-19
Add temperature flag for when generating melodies.

elliotwaite:
Adjusts randomness of the generated melodies by applying:
logits = logits/temperature


magenta/models/attention_rnn/attention_rnn_graph.py
```diff
@@ -153,6 +153,9 @@ def build_graph(mode, hparams_string, input_size, num_classes,
         tf.add_to_collection('summary_op', summary_op)
 
     elif mode == 'generate':
+      if hparams.temperature != 1.0:
+        logits_flat /= hparams.temperature
```
adarob: any reason to not just always do the division?

elliotwaite:
@adarob, to make the graph more efficient by not adding the division op to the graph when not needed.

PR #108, opened 2016-07-25
Dataset pipelines

danabo:
MIDI to dataset transformation pipelines are unified into a common framework. Each conversion from some type to some other type (whether its a protocol buffer or python object) is turned in a modular Pipeline object. Pipelines can be connected together into larger pipelines and run over a dataset.


cghawthorne:
LGTM

PR #109, opened 2016-07-25
Update .gitignore

cghawthorne:
- Ignore all .pyc files
- Switch from specific bazel paths to just /bazel-*. This works better
  because the exact name of the bazel directory changes based on the
  name of the repo directory. For example, if you clone magenta into a
  directory called magenta-work, bazel will create a bazel-magenta-work
  directory.


fredbertsch:
LGTM

Issue #115, opened 2016-07-28
No output about sequenced midi files


hpssjellis:
Anyone else having this issue?

Everything seems to work fine except my output used to give me some indication about the midi files that have been processed, 

I used to see something like:

INFO: found 9 melodies in 3 sequences.

hpssjellis:
Just looked at Tensorboard and I must still have something wrong. The loss is a straight line.

danabo:
The melody info logging went away in monday's change as well. We will bring
it back soon. Sit tight.

hpssjellis:
Last week my starting loss was about 40,000 this week my loss values are about 1.6 ?

danabo:
I don't think we changed anything about the model architectures themselves.
Its possible that the melody dataset being created is different.

danabo:
Regarding the original issue, the *_create_dataset.py scripts will now log how many melodies they extract from how many inputs. Additionally extra information is logged such as why melodies were discarded and statistics about melody lengths.

PR #116, opened 2016-07-28
Add flag for setting logging verbosity.

elliotwaite:
Adds --log flag to callable scripts.
Valid values: DEBUG, INFO (default), WARN, ERROR, or FATAL.

Example usage:

bazel build //magenta/models/basic_rnn:basic_rnn_create_dataset

./bazel-bin/magenta/models/basic_rnn/basic_rnn_create_dataset \
--input=/tmp/notesequences.tfrecord \
--output_dir=/tmp/basic_rnn/sequence_examples \
--eval_ratio=0.10 \
--log=DEBUG


magenta/models/shared/melody_rnn_generate.py
```diff
@@ -147,7 +149,7 @@ def run(melody_encoder_decoder, build_graph):
         (mode, hparams_string, input_size, num_classes, sequence_example_file).
         For an example usage, see models/basic_rnn/basic_rnn_graph.py.
   """
-  tf.logging.set_verbosity(tf.logging.INFO)
+  tf.logging.set_verbosity(FLAGS.log)
```
danabo: set_verbosity takes a string?

danabo:
LGTM.
We will have to remember to do this for every binary. Pretty annoying.

magenta/models/shared/melody_rnn_generate.py
elliotwaite: Yeah, it can take an int, long, or string.

cghawthorne:
LGTM

cghawthorne:
This should fix #115, right?

danabo:
Not quite. The actual logging statements were removed.

danabo:
#115 is fixed. To update this PR, add a `log` flag to this file where tf.logging.set_verbosity is called: https://github.com/tensorflow/magenta/blob/master/magenta/models/shared/melody_rnn_create_dataset.py#L95

PR #117, opened 2016-07-28
Add bpm to command line flags

hpssjellis:
Add bpm default 120 to the shared melody_rnn_generate.py

### DIFF
--- a/magenta/models/shared/melody_rnn_generate.py
+++ b/magenta/models/shared/melody_rnn_generate.py
@@ -60,6 +60,9 @@ tf.app.flags.DEFINE_float('temperature', 1.0,
                           'the unaltered softmax probabilities, greater than '
                           '1.0 makes melodies more random, less than 1.0 makes '
                           'melodies less random.')
+tf.app.flags.DEFINE_integer('bpm', 120,
+                            'Beats per minute for the entire output melody '
+                            'where the default is 120 bpm ')
 
 
 DEFAULT_STEPS_PER_BEAT = 4
@@ -80,6 +83,7 @@ def run_generate(graph, train_dir, output_dir, melody_encoder_decoder,
         generated from scratch.
     num_steps: The total number of steps the final melodies should be,
         priming melody + generated steps. Each step is a 16th of a bar.
+    temperature: Degree of randomness where default=1
     bpm: The tempo in beats per minute that the generated MIDI files will have.
   """
   inputs = graph.get_collection('inputs')[0]
@@ -198,7 +202,7 @@ def run(melody_encoder_decoder, build_graph):
                       FLAGS.primer_midi)
 
   run_generate(graph, train_dir, FLAGS.output_dir, melody_encoder_decoder,
-               primer_melody, FLAGS.num_steps, bpm)
+               primer_melody, FLAGS.num_steps, FLAGS.bpm)
 
 
 if __name__ == '__main__':
--- a/magenta/models/shared/melody_rnn_generate.py
+++ b/magenta/models/shared/melody_rnn_generate.py
@@ -84,7 +84,7 @@ def run_generate(graph, train_dir, output_dir, melody_encoder_decoder,
     num_steps: The total number of steps the final melodies should be,
         priming melody + generated steps. Each step is a 16th of a bar.
     temperature: Degree of randomness where default=1
-    bpm: The tempo in beats per minute that the generated MIDI files will have.
+    bpm: The tempo in beats per minute that the entire generated MIDI files will have. Default = 120
   """
   inputs = graph.get_collection('inputs')[0]
   initial_state = graph.get_collection('initial_state')[0]
--- a/magenta/models/shared/melody_rnn_generate.py
+++ b/magenta/models/shared/melody_rnn_generate.py
@@ -60,9 +60,11 @@ tf.app.flags.DEFINE_float('temperature', 1.0,
                           'the unaltered softmax probabilities, greater than '
                           '1.0 makes melodies more random, less than 1.0 makes '
                           'melodies less random.')
-tf.app.flags.DEFINE_integer('bpm', 120,
-                            'Beats per minute for the entire output melody '
-                            'where the default is 120 bpm ')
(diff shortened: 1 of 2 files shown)


hpssjellis:
I signed it!

magenta/models/shared/melody_rnn_generate.py
```diff
@@ -80,7 +83,8 @@ def run_generate(graph, train_dir, output_dir, melody_encoder_decoder,
         generated from scratch.
     num_steps: The total number of steps the final melodies should be,
         priming melody + generated steps. Each step is a 16th of a bar.
-    bpm: The tempo in beats per minute that the generated MIDI files will have.
+    temperature: Degree of randomness where default=1
```
danabo: temperature needs to be an argument of run_generate

danabo: Actually it looks like temperature is not used in run_generate. You can just take it out of the Args comment.

danabo:
Thanks for making this change. I made one small comment in the code. Please address it and I will merge your PR.

magenta/models/shared/melody_rnn_generate.py
hpssjellis: I have been using temperature so it must be active. I think this line activates it using the hparams

https://github.com/hpssjellis/magenta/blob/master/magenta/models/shared/melody_rnn_generate.py#L171

magenta/models/shared/melody_rnn_generate.py
```diff
@@ -198,7 +202,7 @@ def run(melody_encoder_decoder, build_graph):
                       FLAGS.primer_midi)
 
   run_generate(graph, train_dir, FLAGS.output_dir, melody_encoder_decoder,
-               primer_melody, FLAGS.num_steps, bpm)
+               primer_melody, FLAGS.num_steps, FLAGS.bpm)
```
cghawthorne: We should be careful here. Right now, the bpm is derived from either the BPM present in the priming sequence or a default of 120. This change would remove the ability to base the BPM on the BPM present in the priming sequence.

If we have a flag, I'd prefer for it to take effect up on line 185. It could be referenced instead of melodies_lib.DEFAULT_BEATS_PER_MINUTE. That way, if a BPM is present in the priming sequence, it will be used, but if there isn't a BPM specified, the flag value will be used.

hpssjellis: Your right @cghawthorne. I would like the bpm to replace the primer if it is mentioned in the command line, however my code always ignores the primer bpm. I wil see what I can do to change it.

hpssjellis:
@cghawthorne I have tested it and it works. If bpm is not in the command line then the default behaviour occurs. If bpm is less than or equal to zero then the default behaviour also occurs. If bpm is in the command line and is greater than zero then it controls the output bpm if a primer midi is involved or not.

magenta/models/shared/melody_rnn_generate.py
```diff
@@ -60,6 +60,11 @@
                           'the unaltered softmax probabilities, greater than '
                           '1.0 makes melodies more random, less than 1.0 makes '
                           'melodies less random.')
+tf.app.flags.DEFINE_integer('bpm', -1,
```
cghawthorne: Why not just have a default of 120?

danabo:
This code has changed. It should actually be pretty simple to add a bpm flag in the new code. Please sync your branch to upstream master.

This is where the bpm is set in the new code:
https://github.com/tensorflow/magenta/blob/master/magenta/models/shared/melody_rnn_generate.py#L138
You can have the bpm flag override the bpm value computed here if its a positive value. I don't think we need the ability to set bpm to 0, so you can make that the default instead of -1.

magenta/models/shared/melody_rnn_generate.py
danabo: I think the idea is that this flag will override the bpm in a primer midi file if given.

hpssjellis:
I am just going to close this PR until I have tested it. I see the changes that are needed.

hpssjellis:
@cghawthorne @danabo I think I will wait with my PR until this issue is cleaned up 

https://github.com/tensorflow/magenta/issues/121

The latest change seems to have messed up the auto random note if no primer supplied.

hpssjellis:
@danabo having a hard time figuring out how bpm gets from the new code 

https://github.com/tensorflow/magenta/blob/master/magenta/models/shared/melody_rnn_generate.py#L138

to the sequence generator. Does not seem to be passed as a simple variable at

https://github.com/tensorflow/magenta/blob/master/magenta/models/shared/melody_rnn_generate.py#L167

danabo:
Ah good observation. bpm is not passed into the generator, but is recalculated here
https://github.com/tensorflow/magenta/blob/master/magenta/models/shared/melody_rnn_sequence_generator.py#L97

I think it would be reasonable to pass in bpm into the constructor here
https://github.com/tensorflow/magenta/blob/master/magenta/models/shared/melody_rnn_sequence_generator.py#L31

hpssjellis:
I see hparams is one of the passed variables, in your opinion, should I try what Elliote did at 

https://github.com/tensorflow/magenta/blob/master/magenta/models/shared/melody_rnn_generate.py#L73-L74

>   hparams = ast.literal_eval(FLAGS.hparams if FLAGS.hparams else '{}')
>   hparams['temperature'] = FLAGS.temperature


and make the FLAGS.bpm into an hparam?

danabo:
I think bpm should not be a hyperparameter. Ultimately, a hyperparameter affects what type of model you have, and what it learns. My rule of thumb is that parameters which affect the Tensorflow graph and get saved into checkpoints are hyperparameters. bpm is a parameter about how to convert the model's output into something a human can listen to, and more importantly it depends on the given sample, and not intrinsic to the model itself. Though temperature doesn't affect training it does affect the generated output, but I would accept temperature as an hparam or just a flag.

hpssjellis:
@danabo . Too bad the hparams method was fairly easy. I will look into changing the constructor.

PR #119, opened 2016-07-28
Add SequenceGenerator interface

cghawthorne:
Adds the SequenceGenerator interface and modifies the existing models to use it.

The goal is to expose generator functionality through a common protocol. This change shouldn't modify the functionality of the existing generators, just prepare them for use by other interfaces.


fredbertsch:
LGTM

Issue #121, opened 2016-07-29
No random single note output when primer_midi or primer_melody not supplied


hpssjellis:
Latest change only works if either a primer_midi or primer_melody is supplied

Give this error if neither are supplied, instead of starting the melody with a random note

 

>    bpm = (primer_sequence.tempos[0].bpm if primer_sequence.tempos
> AttributeError: 'NoneType' object has no attribute 'tempos'
>

danabo:
I didn't know that we allowed no primer to be given previously, but you are right. This is the code that picked a random primer:
https://github.com/tensorflow/magenta/blob/4c0050aee5f2bac3f95041a42074cfb6054071c2/magenta/models/shared/melody_rnn_generate.py#L101

We can add this option back in.

hpssjellis:
Sounds good. Any idea of your time frame for the fix?

danabo:
Should be a small change, so early next week seems reasonable.

hpssjellis:
@danabo I am testing a solution to this.

danabo:
I created a PR that will fix both issues - bpm flag and no primer is allowed. This was a bit more complicated than I was anticipated, so I thought it would be easier if I fix both in one PR.

hpssjellis:
Sounds great Daniel. I will close my PR.

hpssjellis:
Solved by @danabo PR 

https://github.com/tensorflow/magenta/pull/128

PR #124, opened 2016-08-01
Re-added a random primer when no melody or midi primer supplied in the command line

hpssjellis:
Address issue https://github.com/tensorflow/magenta/issues/121 recent update removed default behavior that supplied a random primer if a midi file or melody was not supplied in the command line.

### DIFF
--- a/magenta/models/shared/melody_rnn_generate.py
+++ b/magenta/models/shared/melody_rnn_generate.py
@@ -127,7 +127,8 @@ def run_with_flags(melody_rnn_sequence_generator):
     os.makedirs(FLAGS.output_dir)
 
 
-#primer_sequence from written melody or given midi file or random note between midi #48 = 4C to midi #84 = 7C
+# primer_sequence must be set to either: primer_melody or primer_midi or midi #60 = 5C 
+# If set to midi #60 it will be converted to a random note in melody_rnn_sequence_generator.py
 
   primer_sequence = None
   if FLAGS.primer_melody:
@@ -138,8 +139,7 @@ def run_with_flags(melody_rnn_sequence_generator):
     primer_sequence = midi_io.midi_file_to_sequence_proto(FLAGS.primer_midi)
   else:  
     primer_melody = melodies_lib.MonophonicMelody()
-    myRand = random.randint(48,84)
-    primer_melody.from_event_list('['+str(myRand)+']')
+    primer_melody.from_event_list('[60]')
     primer_sequence = primer_melody.to_sequence()
 
   # Derive the total number of seconds to generate based on the BPM of the
--- a/magenta/models/shared/melody_rnn_generate.py
+++ b/magenta/models/shared/melody_rnn_generate.py
@@ -19,6 +19,7 @@ Uses flags to define operation.
 import ast
 import os
 import time
+import random
 
 # internal imports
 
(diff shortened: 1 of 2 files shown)


magenta/models/shared/melody_rnn_generate.py
```diff
@@ -125,13 +126,21 @@ def run_with_flags(melody_rnn_sequence_generator):
   if not os.path.exists(FLAGS.output_dir):
     os.makedirs(FLAGS.output_dir)
 
+
+#primer_sequence from written melody or given midi file or random note between midi #48 = 4C to midi #84 = 7C
```
cghawthorne: @danabo doesn't melody_rnn_sequence_generator already do this?

https://github.com/tensorflow/magenta/blob/master/magenta/models/shared/melody_rnn_sequence_generator.py#L122

danabo: Ah okay so we do still generate a random primer when no primer is given. We just forgot about that case when we compute bpm.

hpssjellis: I see the code, but I don't think it is working. I will test it some more.

hpssjellis: Is it working for you? I get this error when I don't include the primer in the command line

>     bpm = (primer_sequence.tempos[0].bpm if primer_sequence.tempos
> AttributeError: 'NoneType' object has no attribute 'tempos'
> ERROR: Non-zero return code '1' from command: Process exited with status 1.

hpssjellis: My code solves the above error, but I think the original code is generating the random note. There probably is a better, less confusing way to solve the problem.

hpssjellis: @cghawthorne and @danabo  So I changed my code to just reset primer_sequence to ['60'] if no primer_midi or primer_melody. 

https://github.com/hpssjellis/magenta/blob/master/magenta/models/shared/melody_rnn_generate.py#L139-L142

That should be a less confusing method to set the primer_sequence while allowing the original code to change it to a random first note. I will test my code. Perhaps you guys have a better way.

magenta/models/shared/melody_rnn_generate.py
```diff
@@ -125,13 +126,20 @@ def run_with_flags(melody_rnn_sequence_generator):
...
   primer_sequence = None
   if FLAGS.primer_melody:
     primer_melody = melodies_lib.MonophonicMelody()
     primer_melody.from_event_list(ast.literal_eval(FLAGS.primer_melody))
     primer_sequence = primer_melody.to_sequence()
   elif FLAGS.primer_midi:
     primer_sequence = midi_io.midi_file_to_sequence_proto(FLAGS.primer_midi)
+  else:  
+    primer_melody = melodies_lib.MonophonicMelody()
+    primer_melody.from_event_list('[60]')
```
danabo: I thought you wanted to have it randomly sample the first note if no primer is given.

hpssjellis: @danabo @cghawthorne 

It now does. My code only stops the error that was causing the magenta code not to work. When primer_sequence is null the Magenta random primer code fails. Now with primer_sequence set at note 5C the Magenta code replaces it with a random note. 

Probably other ways to do it but this gets it working.

hpssjellis:
@danabo  Good work getting it going. Do you want me to test your code. I had issues with the code dying if bpm was set in the command line 14

I think the issue is at 

https://github.com/tensorflow/magenta/blob/master/magenta/models/shared/melody_rnn_sequence_generator.py#L91

PR #127, opened 2016-08-02
Add sequence diagram for MIDI interface

cghawthorne:


fredbertsch:
Do we want to use the correct MIDI terms? "MIDI keyboard" is oddly specific. I'm not all that familiar with MIDI, but "MIDI Instrument" is (I think) a device that can both send and receive messages.

cghawthorne:
Good call. Done.

fredbertsch:
LGTM

PR #128, opened 2016-08-02
Addresses issue #121

danabo:
Two things are fixed:
1) A bpm flag is added to set the output BPM when the primer is given as
a list of notes or no primer is given.
2) If no primer is given, the first note is randomly sampled.


danabo:
Addresses issue #121

danabo:
@cghawthorne please review.

danabo:
All tests pass. I manually verified that each rnn generator works with --primer_midi, --primer_melody, and no primer given. And I verified that the --bpm flag sets the output bpm when --primer_melody or no primer is given, but not when --primer_midi is given.

hpssjellis:
@danabo Good idea making bpm a float. Hopefully that solves my issue in which bpm <= 14 caused an error.

hpssjellis:
@danabo Well done Daniel:

bpm works great for no primer and primer_melody. bpm is not activated when a primer_midi is called, Not how I would have it, but you might want it that way.

(The reason I wish to control bpm is that temperature messes up bpm and controlling it gets the generated melody back to a sensible bpm. Users can just exclude bpm from the command line if they wish to use the primer_midi bpm for the generated output)

I like the new logging.info. One minor suggestion is that when a random note is generated you output (logging.info() )which note it is.

magenta/models/shared/melody_rnn_generate.py
```diff
@@ -59,6 +59,10 @@
                            'will be used as a priming melody. If a primer '
                            'melody is not specified, melodies will be '
                            'generated from scratch.')
+tf.app.flags.DEFINE_float('bpm', 0,
+                          'The beats per minute to play generated output at. '
+                          'If 0, bpm will default to 120. If the primer has a '
```
adarob: instead of 0, can we make this default at None?

magenta/models/shared/melody_rnn_generate.py
```diff
@@ -59,6 +59,10 @@
...
+                          'bpm that will be used instead.')
```
adarob: it looks like this value is not used if the primer is a MIDI file? can you clarify?

danabo: Clarified.

magenta/models/shared/melody_rnn_generate.py
danabo: Done.

adarob:
LGTM

PR #131, opened 2016-08-04
DAG pipeline

danabo:
This change defines DAGPipeline which runs arbitrary directed acyclic graphs composed of other Pipeline instances. It also defines Statistic which holds useful feedback about pipeline runs which can be aggregated.

melody_rnn pipeline is converted to a DAGPipeline. READMEs are updated with documentation about these new features.


danabo:
@cghawthorne @fredbertsch

cghawthorne:
LGTM

PR #135, opened 2016-08-05
Use a specific checkpoint file for generation

cghawthorne:
This makes it easier to share checkpoint files with others.


danabo:
LGTM

PR #136, opened 2016-08-05
Zero velocity noteon

hanzorama:
Treats an incoming MIDI message with ‘note_on’ status and 0 velocity as a ‘note_off’ message (running status compatibility)


hanzorama:
I signed it!

PR #139, opened 2016-08-05
treat note_on 0 velocity as note_off

hanzorama:
Treats a MIDI message with note_on status and 0 velocity as a note_off message with the same note value (running status compatible)


adarob:
LGTM, thanks!

PR #143, opened 2016-08-06
metronome_channel flag

hanzorama:
-Adds an flag option to define the MIDI channel the metronome messages
come in on (comes in on MIDI Channel 1 by default). Allows separating
generated content and click (for recording/monitoring)

-Ability to add metronome to playback of generated material

-Adds a flag option to define the mix amount of metronome click that is
heard during playback vs capture. By default, it is set to 0% so that
no click is heard unless flag is set.

-include default metronome velocity as a variable

### DIFF
--- a/magenta/interfaces/midi/midi.py
+++ b/magenta/interfaces/midi/midi.py
@@ -80,6 +80,15 @@ tf.app.flags.DEFINE_integer(
     'num_bars_to_generate',
     5,
     'The number of bars to generate each time.')
+tf.app.flags.DEFINE_integer(
+    'metronome_channel',
+    0,
+    'The MIDI channel on which to send the metronome click')
+tf.app.flags.DEFINE_integer(
+    'metronome_playback_mix',
+    0,
+    'An integer between 0 and 100 expressing the percentage of '
+    'metronome click to be heard during generated playback')
 tf.app.flags.DEFINE_string(
     'generator_name',
     None,
@@ -104,6 +113,8 @@ _GENERATOR_FACTORY_MAP = {
 
 _METRONOME_TICK_DURATION = 0.05
 _METRONOME_PITCH = 95
+# TODO(hanzorama): Make velocity adjustable by a control change signal.
+_METRONOME_VELOCITY = 64
 
 
 def serialized(func):
@@ -198,11 +209,12 @@ class Metronome(threading.Thread):
   """
   daemon = True
 
-  def __init__(self, outport, bpm, clock_start_time):
+  def __init__(self, outport, bpm, clock_start_time, velocity):
     self._outport = outport
     self._bpm = bpm
     self._stop_metronome = False
     self._clock_start_time = clock_start_time
+    self._metronome_velocity = int(round(velocity))
     super(Metronome, self).__init__()
 
   def run(self):
@@ -230,9 +242,14 @@ class Metronome(threading.Thread):
         while time.time() < next_tick_time:
           pass
 
-      self._outport.send(mido.Message(type='note_on', note=_METRONOME_PITCH))
+      self._outport.send(
+          mido.Message(
+              type='note_on', note=_METRONOME_PITCH, channel=FLAGS.metronome_channel,
+              velocity=self._metronome_velocity))
       time.sleep(_METRONOME_TICK_DURATION)
-      self._outport.send(mido.Message(type='note_off', note=_METRONOME_PITCH))
+      self._outport.send(
+          mido.Message(
+              type='note_off', note=_METRONOME_PITCH, channel=FLAGS.metronome_channel))
 
   def stop(self):
     """Signals for the metronome to stop and joins thread."""
@@ -247,25 +264,34 @@ class MonoMidiPlayer(threading.Thread):
     _outport: The Mido port for sending messages.
     _sequence: The monohponic, chronologically sorted NoteSequence to play.
     _stop_playback: A boolean specifying whether the playback should stop.
+    _bpm: The tempo
   Args:
     outport: The Mido port for sending messages.
     sequence: The monohponic, chronologically sorted NoteSequence to play.
   """
   daemon = True
 
+  if not 0 <= FLAGS.metronome_playback_mix <= 100:
+      raise ValueError(
+          'The metronome_playback_mix must be between 0 and 100')
+
   def __init__(self, outport, sequence):
     self._outport = outport
     self._sequence = sequence
     self._stop_playback = False
+    self._bpm = sequence.tempos[0].bpm
+    self._metronome = Metronome(self._outport,self._bpm, time.time(),
+                                _METRONOME_VELOCITY*FLAGS.metronome_playback_mix/100)
     super(MonoMidiPlayer, self).__init__()
 
   def run(self):
-    """Plays back the NoteSequence until it ends or stop signal is received.
+    """Plays back the NoteSequence with metronome until it ends or stop signal is received.
 
     Raises:
       ValueError: The NoteSequence is not monophonic and chronologically sorted.
     """
     stdout_write_and_flush('Playing sequence...')
+    self._metronome.start()
     # Wall start time.
     play_start = time.time()
     # Time relative to start of NoteSequence.
@@ -298,11 +324,13 @@ class MonoMidiPlayer(threading.Thread):
       if delta > 0:
         time.sleep(delta)
       self._outport.send(mido.Message('note_off', note=note.pitch))
+    self._metronome.stop()
     stdout_write_and_flush('Done\n')
 
   def stop(self):
-    """Signals for the playback to stop and joins thread."""
+    """Signals for the playback and metronome to stop and joins thread."""
     self._stop_playback = True
+    self._metronome.stop()
     self.join()
 
 
@@ -419,7 +447,7 @@ class MonoMidiHub(object):
     self._sequence_start_time = None
     self._capture_start_time = time.time()
     self._inport.callback = self._capture_message
-    self._metronome = Metronome(self._outport, bpm, self._capture_start_time)
+    self._metronome = Metronome(self._outport, bpm, self._capture_start_time,_METRONOME_VELOCITY)
     self._metronome.start()
 
   @serialized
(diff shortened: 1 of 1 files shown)


magenta/interfaces/midi/midi.py
```diff
@@ -230,9 +242,14 @@ def run(self):
         while time.time() < next_tick_time:
           pass
 
-      self._outport.send(mido.Message(type='note_on', note=_METRONOME_PITCH))
+      self._outport.send(
+          mido.Message(
+              type='note_on', note=_METRONOME_PITCH, channel=FLAGS.metronome_channel,
```
adarob: let's avoid using flags outside of the main function

magenta/interfaces/midi/midi.py
```diff
@@ -247,25 +264,34 @@ class MonoMidiPlayer(threading.Thread):
...
     _sequence: The monohponic, chronologically sorted NoteSequence to play.
     _stop_playback: A boolean specifying whether the playback should stop.
+    _bpm: The tempo
   Args:
     outport: The Mido port for sending messages.
     sequence: The monohponic, chronologically sorted NoteSequence to play.
   """
   daemon = True
 
+  if not 0 <= FLAGS.metronome_playback_mix <= 100:
```
adarob: do this check in main

magenta/interfaces/midi/midi.py
```diff
@@ -247,25 +264,34 @@ class MonoMidiPlayer(threading.Thread):
     _outport: The Mido port for sending messages.
     _sequence: The monohponic, chronologically sorted NoteSequence to play.
     _stop_playback: A boolean specifying whether the playback should stop.
+    _bpm: The tempo
```
adarob: The tempo in beats per minute. Copied from input sequence.

magenta/interfaces/midi/midi.py
```diff
@@ -247,25 +264,34 @@ class MonoMidiPlayer(threading.Thread):
...
 
+  if not 0 <= FLAGS.metronome_playback_mix <= 100:
+      raise ValueError(
+          'The metronome_playback_mix must be between 0 and 100')
+
   def __init__(self, outport, sequence):
     self._outport = outport
     self._sequence = sequence
     self._stop_playback = False
+    self._bpm = sequence.tempos[0].bpm
```
adarob: Let's add a check that there is exactly 1 tempo.

adarob: Actually, you don't really need to store the bpm as an attribute, do you?

magenta/interfaces/midi/midi.py
```diff
@@ -247,25 +264,34 @@ class MonoMidiPlayer(threading.Thread):
...
+    self._metronome = Metronome(self._outport,self._bpm, time.time(),
```
adarob: add space after "self._outport,"

magenta/interfaces/midi/midi.py
```diff
@@ -247,25 +264,34 @@ class MonoMidiPlayer(threading.Thread):
...
+                                _METRONOME_VELOCITY*FLAGS.metronome_playback_mix/100)
```
adarob: add spaces between all symbols

magenta/interfaces/midi/midi.py
```diff
@@ -247,25 +264,34 @@ class MonoMidiPlayer(threading.Thread):
...
     super(MonoMidiPlayer, self).__init__()
 
   def run(self):
-    """Plays back the NoteSequence until it ends or stop signal is received.
+    """Plays back the NoteSequence with metronome until it ends or stop signal is received.
```
adarob: Lines should be at most 80 characters (sorry for these nit-picky google style rules)

magenta/interfaces/midi/midi.py
```diff
@@ -80,6 +80,15 @@
     'num_bars_to_generate',
     5,
     'The number of bars to generate each time.')
+tf.app.flags.DEFINE_integer(
+    'metronome_channel',
+    0,
+    'The MIDI channel on which to send the metronome click')
```
adarob: end description with period

magenta/interfaces/midi/midi.py
```diff
@@ -80,6 +80,15 @@
...
+tf.app.flags.DEFINE_integer(
+    'metronome_playback_mix',
+    0,
+    'An integer between 0 and 100 expressing the percentage of '
+    'metronome click to be heard during generated playback')
```
adarob: end with period

adarob:
Sorry for all of these nit-picky google style rules!

adarob:
FYI: https://google.github.io/styleguide/pyguide.html

hanzorama:
Great, this is a good thing to know about ;) No need to apologize!

magenta/interfaces/midi/midi.py
hanzorama: nope, you're right. Also, if there are multiple tempos is it only because of a tempo change during the capture of the NoteSequence? And would sequence.tempos[-1].bpm be the latest tempo or the starting tempo (or something else)?

hanzorama:
I still need to add the sequence.tempos check

magenta/interfaces/midi/midi.py
```diff
@@ -198,11 +209,13 @@ class Metronome(threading.Thread):
   """
   daemon = True
 
-  def __init__(self, outport, bpm, clock_start_time):
+  def __init__(self, outport, bpm, clock_start_time, velocity, channel):
     self._outport = outport
     self._bpm = bpm
     self._stop_metronome = False
     self._clock_start_time = clock_start_time
+    self._metronome_velocity = int(round(velocity))
```
adarob: Let's just call these _velocity and _channel for simplicity since it's clear they relate to the metronome.

magenta/interfaces/midi/midi.py
```diff
@@ -247,16 +263,23 @@ class MonoMidiPlayer(threading.Thread):
...
     _stop_playback: A boolean specifying whether the playback should stop.
+    _bpm: The tempo in beats per minute. Copied from input sequence.
   Args:
     outport: The Mido port for sending messages.
     sequence: The monohponic, chronologically sorted NoteSequence to play.
   """
   daemon = True
 
-  def __init__(self, outport, sequence):
+  def __init__(self, outport, sequence, metronome_playback_velocity):
```
adarob: how about just "metronome_velocity" or "velocity"

magenta/interfaces/midi/midi.py
```diff
@@ -510,6 +539,11 @@ def main(unused_argv):
           '--stop_capture_control_value must both be defined and unique.')
     return
 
+  if not (0 <= FLAGS.metronome_playback_velocity <= 127
+          and type(FLAGS.metronome_playback_velocity) is int):
```
adarob: no need to check if it is an int because the flag is defined as an integer

magenta/interfaces/midi/midi.py
```diff
@@ -104,6 +113,8 @@
 
 _METRONOME_TICK_DURATION = 0.05
 _METRONOME_PITCH = 95
+# TODO(hanzorama): Make velocity adjustable by a control change signal.
+_METRONOME_VELOCITY = 64
```
adarob: _METRONOME_CAPTURE_VELOCITY

magenta/interfaces/midi/midi.py
```diff
@@ -198,11 +209,12 @@ class Metronome(threading.Thread):
   """
   daemon = True
 
-  def __init__(self, outport, bpm, clock_start_time):
+  def __init__(self, outport, bpm, clock_start_time, velocity):
     self._outport = outport
     self._bpm = bpm
     self._stop_metronome = False
     self._clock_start_time = clock_start_time
+    self._velocity = int(round(velocity))
```
adarob: Let's just assume this is already an integer.

magenta/interfaces/midi/midi.py
```diff
@@ -198,11 +209,12 @@ class Metronome(threading.Thread):
   """
```
adarob: Update attributes and args in docstring

magenta/interfaces/midi/midi.py
```diff
@@ -247,16 +262,21 @@ class MonoMidiPlayer(threading.Thread):
...
     sequence: The monohponic, chronologically sorted NoteSequence to play.
   """
   daemon = True
 
-  def __init__(self, outport, sequence):
+  def __init__(self, outport, sequence, metronome_velocity):
     self._outport = outport
     self._sequence = sequence
     self._stop_playback = False
+    if len(sequence.tempos) == 1:
```
adarob: Raise an exception if there isn't a single tempo. Make sure to document that exception in the docstring for the class.

adarob: preferred:
if not len(sequence.tempos) == 1:
  raise ...
bpm = ...

magenta/interfaces/midi/midi.py
```diff
@@ -247,16 +264,25 @@ class MonoMidiPlayer(threading.Thread):
     _outport: The Mido port for sending messages.
     _sequence: The monohponic, chronologically sorted NoteSequence to play.
     _stop_playback: A boolean specifying whether the playback should stop.
+    _bpm: The tempo in beats per minute. Copied from input sequence.
   Args:
     outport: The Mido port for sending messages.
     sequence: The monohponic, chronologically sorted NoteSequence to play.
```
douglaseck: s/monohponic/monophonic
Incidentally I don't think monophonic is really the right word. Single voice? But it gets the point across.

magenta/interfaces/midi/midi.py
```diff
@@ -510,6 +539,9 @@ def main(unused_argv):
           '--stop_capture_control_value must both be defined and unique.')
     return
 
+  if not 0 <= FLAGS.metronome_playback_velocity <= 127:
+    raise ValueError('The velocity must be an integer between 0 and 127')
```
adarob: Let's actually do a print and return like we do above since this is the main function.

magenta/interfaces/midi/midi.py
```diff
@@ -510,6 +538,10 @@ def main(unused_argv):
           '--stop_capture_control_value must both be defined and unique.')
     return
 
+  if not 0 <= FLAGS.metronome_playback_velocity <= 127:
+    print 'The velocity must be an integer between 0 and 127'
```
adarob: Be more specific please: '---metronome_playback_velocity must be...'

adarob:
LGTM, but can you sync with master to get rid of merge conflicts?

PR #146, opened 2016-08-06
Simplifying binary counters + making them consistent between models

eraoul:
Also noticed a missing "len(melody)-1" in the attention model.

### DIFF
--- a/magenta/models/attention_rnn/attention_rnn_encoder_decoder.py
+++ b/magenta/models/attention_rnn/attention_rnn_encoder_decoder.py
@@ -132,8 +132,8 @@ class MelodyEncoderDecoder(melodies_lib.MelodyEncoderDecoder):
         melody.events[-1] == melody.events[-(2 * STEPS_PER_BAR + 1)]):
       input_[self.note_range + 5] = 1.0
 
-    # Binary time counters.
-    n = len(melody) - 1
+    # Binary time counter giving the metric location of the *next* note.
+    n = len(melody)
     for i in range(NUM_BINARY_TIME_COUNTERS):
       input_[self.note_range + 6 + i] = 1.0 if (n / 2 ** i) % 2 else -1.0
 
--- a/magenta/models/attention_rnn/attention_rnn_encoder_decoder.py
+++ b/magenta/models/attention_rnn/attention_rnn_encoder_decoder.py
@@ -24,6 +24,7 @@ NO_EVENT = melodies_lib.NO_EVENT
 MIN_MIDI_PITCH = melodies_lib.MIN_MIDI_PITCH
 NOTES_PER_OCTAVE = melodies_lib.NOTES_PER_OCTAVE
 STEPS_PER_BAR = 16  # This code assumes the melodies have 16 steps per bar.
+
 MIN_NOTE = 48  # Inclusive
 MAX_NOTE = 84  # Exclusive
 TRANSPOSE_TO_KEY = 0  # C Major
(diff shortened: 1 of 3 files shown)


cghawthorne:
@elliotwaite can you take a look at this?

eraoul:
I signed the CLA.

elliotwaite:
Actually, I think it should be "n = len(melody)" for both. It shifts the binary clock forward one to give the correct signal for what the next note should be, instead of signalling what the time position of the last note was. For example, the bar binary counter should be -1.0 for the first 15 inputs, but change to 1.0 for the 16th input, so that it will signal the start of a new bar for generating the 17th step.

For example, instead of n = len(melody) - 1:
1: [0, 0, 0, 0, 0]
2: [1, 0, 0, 0, 0]
3: [0, 1, 0, 0, 0]
4: [1, 1, 0, 0, 0]
5: [0, 0, 1, 0, 0]
6: [1, 0, 1, 0, 0]
7: [0, 1, 1, 0, 0]
8: [1, 1, 1, 0, 0]
9: [0, 0, 0, 1, 0]
10: [1, 0, 0, 1, 0]
11: [0, 1, 0, 1, 0]
12: [1, 1, 0, 1, 0]
13: [0, 0, 1, 1, 0]
14: [1, 0, 1, 1, 0]
15: [0, 1, 1, 1, 0]
16: [1, 1, 1, 1, 0]
17: [0, 0, 0, 0, 1]
18: [1, 0, 0, 0, 1]
19: [0, 1, 0, 0, 1]
20: [1, 1, 0, 0, 1]
21: [0, 0, 1, 0, 1]
22: [1, 0, 1, 0, 1]
23: [0, 1, 1, 0, 1]
24: [1, 1, 1, 0, 1]
25: [0, 0, 0, 1, 1]
26: [1, 0, 0, 1, 1]
27: [0, 1, 0, 1, 1]
28: [1, 1, 0, 1, 1]
29: [0, 0, 1, 1, 1]
30: [1, 0, 1, 1, 1]
31: [0, 1, 1, 1, 1]
32: [1, 1, 1, 1, 1]

It should be n = len(melody), since the input is for predicting the next note:
1: [1, 0, 0, 0, 0]
2: [0, 1, 0, 0, 0]
3: [1, 1, 0, 0, 0]
4: [0, 0, 1, 0, 0]
5: [1, 0, 1, 0, 0]
6: [0, 1, 1, 0, 0]
7: [1, 1, 1, 0, 0]
8: [0, 0, 0, 1, 0]
9: [1, 0, 0, 1, 0]
10: [0, 1, 0, 1, 0]
11: [1, 1, 0, 1, 0]
12: [0, 0, 1, 1, 0]
13: [1, 0, 1, 1, 0]
14: [0, 1, 1, 1, 0]
15: [1, 1, 1, 1, 0]
16: [0, 0, 0, 0, 1]
17: [1, 0, 0, 0, 1]
18: [0, 1, 0, 0, 1]
19: [1, 1, 0, 0, 1]
20: [0, 0, 1, 0, 1]
21: [1, 0, 1, 0, 1]
22: [0, 1, 1, 0, 1]
23: [1, 1, 1, 0, 1]
24: [0, 0, 0, 1, 1]
25: [1, 0, 0, 1, 1]
26: [0, 1, 0, 1, 1]
27: [1, 1, 0, 1, 1]
28: [0, 0, 1, 1, 1]
29: [1, 0, 1, 1, 1]
30: [0, 1, 1, 1, 1]
31: [1, 1, 1, 1, 1]
32: [0, 0, 0, 0, 0]

It's counter intuitive since you're feeding in information about the previous note, but the time clock is the signal for the next note, but I think it's the right strategy. Thanks for finding that inconsistency between lookback_rnn and attention_rnn.

Other than that, the change looks good to me.

adarob:
LGTM, thanks!

eraoul:
Makes sense -- I could see an argument for doing it either way, but yes, I've modified to point at the next note.

An aside: I think the previous (and current) code has "-1" and "1" reversed from what one would expect (i.e., "result = 1 if ....... %2 else -1" seems backwards from what you described above). But it doesn't really matter as long as all the examples are consistent :) Should we swap them or leave it alone?

PR #148, opened 2016-08-09
Fix is sequence

eraoul:
_is_sequence and _unpacked_state disappeared in TF 0.10; I had to update to use tensorflow.python.util.nest instead. I assume this change would break users with TF 0.9, however.


adarob:
Yes, this is unfortunate since it will break 0.9 users. Can you also include an update to the installation instructions to say that 0.10 or greater is required?

eraoul:
Sure thing -- done.

adarob:
Thanks. I'm going to discuss with the team how best to roll this out since it is not backwards compatible.

eraoul:
SGTM. Maybe it's reasonable to add a check to see if those functions exist
to preserve compatibility for a version or two..

PR #151, opened 2016-08-09
Update MonophonicMelody docstring

adarob:


danabo:
Wow I should have caught that when I changed it. Thanks for fixing :)

PR #153, opened 2016-08-11
Don't crash pipeline due to sequence with non-integer steps per bar.

adarob:


danabo:
LGTM.

It might also be useful to record steps_per_bar_float in a histogram statistic so users can figure out what the offending time signature is (or change their steps per beat). Or you can have counters for every time signature.

adarob:
I'm logging a warning that will tell what the bad time signature is.

PR #173, opened 2016-08-22
Fixes and refactors for melody_encoder_decoders

adarob:
-Modify melody encoder/decoder code to avoid destroying encoded/decoded melodies while also avoiding unnecessary copies. 
-Adds  optional NO_EVENT padding to end of extracted melodies to complete bars.
-Adds max_steps_truncate and max_argument_discard arguments for melody extractor that shorten/discard melodies that are too long.
-Removes final NOTE_OFF events from extracted melodies.
-Refactors some melody encoder/decoder code for re-use.


danabo:
LGTM

PR #174, opened 2016-08-24
End melody end note

adarob:
Modifies MonophonicMelody.to_sequence to treat the end of the melody as a NOTE_OFF event. Adds tests for to_sequence. Makes class attributes private.


danabo:
LGTM

PR #178, opened 2016-08-27
Use tf.gfile instead of os.path in sequence_generator.py for internal compatibility.

adarob:


magenta/lib/sequence_generator.py
```diff
@@ -89,14 +88,18 @@ def initialize(self):
...
+      if not tf.gfile.Exists(self._checkpoint):
         raise SequenceGeneratorException(
             'Checkpoint path does not exist: %s' % (self._checkpoint))
       checkpoint_file = self._checkpoint
       # If this is a directory, try to determine the latest checkpoint in it.
-      if os.path.isdir(checkpoint_file):
+      if tf.gfile.IsDirectory(checkpoint_file):
         checkpoint_file = tf.train.latest_checkpoint(checkpoint_file)
-      if not os.path.isfile(checkpoint_file):
+      if checkpoint_file is None:
```
cghawthorne: shouldn't this happen before the isDirectory check?

adarob: latest_checkpoint returns None if none are found in the directory, which I discovered when testing this :)

cghawthorne:
LGTM

Issue #158, opened 2016-08-14
Just cannot create SequenceExamples


npuichigo:
When I want to create SequenceExamples from notesequences.tfrecord, message like this occurs:

`INFO:tensorflow:Processed 1 inputs total. Produced 0 outputs.`

Besides, the SequenceExample files' sizes are zero.

![qq 20160814174942](https://cloud.githubusercontent.com/assets/11533479/17648429/c15bd896-6247-11e6-84aa-2c90fafeca0a.png)

This is my notesequences.tfrecord file created from bach's Goldberg Variations:

![qq 20160814175316](https://cloud.githubusercontent.com/assets/11533479/17648437/0251fb14-6248-11e6-831c-ed8ba0322a59.png)

prafiles:
Are you sure the input sequences are not polyphonic? I'm also encountering similar issue, however if you're working on the latest pull, you'll notice that it does spit the reason out in more verbose explanations.

npuichigo:
Thanks for your reminding, but the README.md in basic_rnn says:
> This model takes monophonic melodies, meaning one note plays at a time. Use basic_rnn_create_dataset.py to extract monophonic melodies from NoteSequence protos made from your MIDI files. The script will look at each instrument track and extract a melody line if it is at least 7 measures long, and at least 5 unique pitches (with octave equivalence). If multiple notes play at the same time, one note is kept.

Does that mean polyphonic input is okay? And monophonic melodies can be extracted from original polyphonic input?

prafiles:
I've seen the "extraction mechanism" give strange outputs, to nothing. Unfortunately I'm away from my workstation, but can share the complete log with the latest pull. Both while creating the tfrecord files and then datasets?

npuichigo:
While creating the tfrecord files and datasets, messages are printed out respectively as follows:
> zhangyuchao@node1-cluster:~/tf_workspace/magenta$./dataset_processing.sh
INFO: Found 1 target...
Target //magenta/scripts:convert_midi_dir_to_note_sequences up-to-date:
  bazel-bin/magenta/scripts/convert_midi_dir_to_note_sequences
INFO: Elapsed time: 0.119s, Critical Path: 0.00s
[...]

PR #181, opened 2016-08-30
Quantizer counts sequences with multiple time signatures

danabo:
Response to #158. When sequences are discarded because there is more than one time signature the reason will be printed in the terminal.


cghawthorne:
LGTM

PR #183, opened 2016-08-30
Sequence Generator Bundles

cghawthorne:
Add the ability to create and read sequence generator bundle files. Bundles contain the checkpoint, metagraph, and generator details and can serve as standalone representations of generator state.


adarob:
LGTM

PR #190, opened 2016-09-01
Switch to using QPM instead of BPM

cghawthorne:
QPM (quarter notes per minute) is much less ambiguous than BPM (beats per minute). Quarter notes are a clearly defined entity, whereas the meaning of "beat" could change based on the time signature (e.g., in 2/2, the half note gets the beat) or could even be considered to be based on the listener's perception of where the beat happens. QPM also matches how MIDI and MusicXML files are encoded.

Our existing code was actually written assuming the tempo was indicated in QPM, this rename just makes it more clear.


adarob:
LGTM

PR #191, opened 2016-09-01
Add output_file_base arg to pipeline runner.

adarob:


danabo:
Is this change being used anywhere?

adarob:
I'm using it internally to make it equivalent to the MR version.

danabo:
Ah, I see. LGTM

PR #192, opened 2016-09-02
pretty_music21

czhuang:
A preliminary pretty_midi-like read-only interface to music21.  Can be used to convert musicXML files into NoteSequence proto.


czhuang:
I signed it!

cghawthorne:
LGTM

PR #193, opened 2016-09-02
Add copy method to MonophonicMelody.

adarob:


magenta/lib/melodies_lib.py
```diff
@@ -248,6 +254,10 @@ def _get_last_on_off_events(self):
         return (i, last_off)
     raise ValueError('No events in the stream')
 
+  def copy(self):
+    """Returns a deep copy of this MonphonicMelody object."""
```
cghawthorne: nit: probably shouldn't call this a "deep" copy because it uses the shallow copy operator (__copy__). I think in practice it does end up being a deep copy, but that's only because from_event_list happens to use an iterator over a list of ints.

adarob: Yeah I went back and forth on this but what we really need is a deep copy
so that we aren't accidentally modifying the state of the original object.
Should we just name the functions deepcopy?

cghawthorne: Yeah, if that's what we want, that's what we should call it.

Also, might not hurt to explicitly make a copy of the events list, just to be clear.

adarob:
Renamed. PTAL.

magenta/lib/melodies_lib.py
```diff
@@ -248,6 +248,15 @@ def _get_last_on_off_events(self):
         return (i, last_off)
     raise ValueError('No events in the stream')
 
+  def deepcopy(self):
```
cghawthorne: might as well also define __deecopy__ while you're here, in case someone wants to use copy.deepcopy

adarob: Done.

cghawthorne:
LGTM

PR #196, opened 2016-09-02
check for oversized time signature denominator

iansimon:
Check for bad time signature denominator values in MIDI.

MIDI represents the denominator of a time signature as 2 raised to a one-byte value.  As such, it is possible, though uncommon and likely due to corrupt data, for the denominator to exceed the maximum value of an int32.  In this case, raise a MIDIConversionError.


adarob:
LGTM

PR #203, opened 2016-09-07
Fix CheckReadWriteMidi so that it sanitizes MIDI data.

douglaseck:
This is done to ensure correct ordering. Add a test to check specifically that this ordering of events is correctly done. (Before the fix in this CL for CheckReadWriteMidi, testEventOrdering fails). Uncomment  testComplexReadWriteMidi, as this CL fixes the test. Update comments throughout.


cghawthorne:
LGTM

PR #209, opened 2016-09-12
CC Map and interface

hanzorama:
Allows user definition for start/stop capture messages, metronome velocity changes via CC, and changes structure to allow easy addition of other user definable CC parameters.


magenta/interfaces/midi/midi.py
```diff
@@ -498,48 +516,216 @@ def stop_capture(self):
...
+    Args:
+      message: The incoming Control Change message.
+    """
+
+    # Metronome Velocity
+    if message.bytes()[:2] == _CC_MAP['Metronome velocity'].bytes()[:2]:
+      global _METRONOME_VELOCITY
+      _METRONOME_VELOCITY = message.bytes()[2]
+
+class CCRemapper(object):
```
adarob: Let's start the modularization of this file by moving this remapper to its own file.

magenta/interfaces/midi/midi.py
```diff
@@ -119,10 +123,16 @@
...
-# TODO(hanzorama): Make velocity adjustable by a control change signal.
-_METRONOME_CAPTURE_VELOCITY = 64
+_METRONOME_VELOCITY = 64
 
 
+_CC_MAP = {
+  'Start Capture': mido.Message('control_change', channel=0, control=1,
+                                value=127),
+  'Stop Capture': mido.Message('control_change', channel=0, control=1, value=0),
+  'Metronome velocity': mido.Message('control_change', channel=0, control=3),
```
adarob: Capitalize 'Velocity' for consistency.

magenta/interfaces/midi/midi.py
```diff
@@ -498,48 +516,216 @@ def stop_capture(self):
...
+  def execute_cc_message(self, message):
+    """Defines how to treat non Start/Stop user defined CC messages.
+
+    Args:
+      message: The incoming Control Change message.
+    """
+
+    # Metronome Velocity
+    if message.bytes()[:2] == _CC_MAP['Metronome velocity'].bytes()[:2]:
+      global _METRONOME_VELOCITY
```
adarob: Let's not use globals. Instead you can have an object that gets passed around that has the current values of these variables. We can discuss offliine.

magenta/interfaces/midi/midi.py
```diff
@@ -498,48 +516,216 @@ def stop_capture(self):
...
+        print ("{} has been assigned to controller "
+               "number {}".format(cc_choice, msg.control))
+        self._cc_map[cc_choice] = msg
+        return
+      else:
+        print('Sorry, I only accept MIDI CC messages for this parameter..')
+        time.sleep(.5)
+        continue
+
+  def remap_capture_message(self, input_port, cc_choice, msg_start_capture,
```
adarob: Why do we need something specific for the capture message?

magenta/interfaces/midi/midi.py
```diff
@@ -498,48 +516,216 @@ def stop_capture(self):
...
+      print "\nWhat would you like to assign a Control Change to...?:"
+      print "0. None (Exit)"
+      for i, CCparam in enumerate(sorted(self._cc_map.keys())):
+        print "{}. {}".format(i + 1, CCparam)
+      print ""
+      try:
+        cc_choice_index = int(raw_input("Please make a selection...\n>>>"))
+      except ValueError:
+        print "Please enter a number..."
+        time.sleep(1)
```
adarob: why sleep?

magenta/interfaces/midi/midi.py
hanzorama: It is the only CC that has a toggleability (for use with basic foot switches, notes on the keyboard, or MIDI buttons that only send out 1 message). So it needs a separate set of assignment definitions

magenta/interfaces/midi/midi.py
hanzorama: because otherwise a lot of readable info goes by quickly. I took it out and if it bothersome without the pauses we can put it back in.

hanzorama: nevermind, I already miss em. I’ll put them back in and if I really need to take them out I can

adarob:
@hanzorama , I still want to add this feature, but I think it might be worth rethinking a bit now that a lot of the code has changed significantly. We can also simplify things quite a bit now. You'll definitely need to merge with master or start a new branch.

I'm happy to take this on if you don't have time, but if you want to do it, it's all yours!
 
Here is a basic plan:
1. Add a small class that takes a dictionary mapping a string name to a cc number or None and returns an updated version of that dictionary. The class has a method that will enter a user feedback loop that prints the current mapping and allows the user to enter a number to update one of the values or quit. I know you already have something like this now, but I just want to simplify and clean it up a bit.

2. In magenta_midi, iterate through FLAGS, selecting those that end in _control_number and adding them to a dictionary. Before initializing the interaction, optionally pass this dictionary into the remap class based on user input (another flag? or maybe tell the user to press 'c' to map control changes or enter to start the interaction). After remapping, pass these values into the interaction and continue as usual.

Let me know if you have any Q's.

adarob:
Obsolete. Replaced by #526.

PR #210, opened 2016-09-12
Add description field to generator bundle.

adarob:


magenta/protobuf/generator.proto
```diff
@@ -65,10 +65,15 @@ message GenerateSequenceResponse {
...
 // Bundle for wrapping a generator id, checkpoint file, and metagraph.
 // Intended to make sharing pre-trained models easier.
+// Next ID: 5
 message GeneratorBundle {
   // Details about the generator that created this bundle.
   GeneratorDetails generator_details = 1;
 
+  // A short text description of the bundle (e.g., training data, hparams,
+  // etc.).
+  string description = 4;
```
cghawthorne: I think i'd prefer we make a sub-message for bundle-specific metadata. I could see wanting to add other fields (e.g., author) at some point.

magenta/lib/sequence_generator.py
```diff
@@ -214,13 +222,15 @@ def generate(self, generate_sequence_request):
     self.initialize()
     return self._generate(generate_sequence_request)
 
-  def create_bundle_file(self, bundle_file):
+  def create_bundle_file(self, bundle_file, description):
```
cghawthorne: should this have a default value of '' to avoid breaking existing callers?

adarob: Done.

adarob:
PTAL

cghawthorne:
LGTM

PR #214, opened 2016-09-15
MusicXML Parser for Magenta

jsawruk:
Lightweight MusicXML parser for Magenta with no dependencies. Handles uncompressed (.xml) and compressed (.mxl) files. Translates MusicXML file into NoteSequences similar to how MIDI is imported. 

Because MusicXML differs from MIDI, different information is available to Magenta. For example, MusicXML does not contain MIDI CC data. However, MusicXML does contain other information (such as dynamic markings) which may be more useful in certain scenarios, such as using Magenta to generate sheet music (MusicXML files) in addition to MIDI files.

This parser currently only supports input and does not support output to MusicXML files.

Four public domain MusicXML files are included for unit testing.


adarob:
@cghawthorne or @iansimon: can one of your take this review?

cghawthorne:
Yup, I can definitely look at it. Wouldn't mind a second pair of eyes, though.

@jsawruk we'll start the review once the CLA is signed. Thanks again for your work on this!

iansimon:
I'll take a look also.

-Ian

jsawruk:
I signed it!

cghawthorne:
@jsawruk the cla bot still doesn't seem happy. Can you try replying using the github web interface?

jsawruk:
I updated the contact info email address on the CLA to match the email address on my Github account. Hopefully this resolves the issue.

cghawthorne:
Might need to re-reply with "I signed it!"

jsawruk:
I signed it!

willnorris:
@jsawruk: sorry about the confusion here.  It looks like we're not able to map your GitHub username to the email address used in your commits.  Could you add your *@jwpepper.com email address [to your GitHub account](https://help.github.com/articles/adding-an-email-address-to-your-github-account/)?  That should take care of this.

jsawruk:
Ok, I have added that email address to my Github account and it is now verified.

jsawruk:
In process of updating my code. I just rebased against master, and am now getting this error when running the unit tests:
```
  File "musicxml_parser_test.py", line 20, in <module>
    from musicxml_reader import *
  File "/Users/jsawruk/magenta/magenta/music/musicxml_reader.py", line 29, in <module>
    from magenta.protobuf import music_pb2
ImportError: No module named magenta.protobuf
```

What changed?

cghawthorne:
I think you're running into a merge conflict with #223. We recently moved a bunch of our modules that used to be in /lib to /music. You'll probably also want your musicxml code to be in /music now.

jsawruk:
I'm still getting errors. I checked out the master branch at commit 0056294, which includes #223. When I try to import using python, I get this error:

```
>>> from magenta.protobuf import music_pb2
  File "magenta/__init__.py", line 26, in <module>
    import magenta.music.melodies_lib
ImportError: No module named music.melodies_lib
```

This does include the new /music directory, and the BUILD file there looks correct.

iansimon:
Are you having issues when running bazel test, or only in interactive python?  When working on libraries for core magenta, we recommend testing via bazel.

jsawruk:
I wasn't running the bazel test correctly, so I tried it in interactive Python. Once I ran the bazel test correctly, I was able to resolve the issue; the bazel tests now pass on my system.

magenta/music/musicxml_parser.py
```diff
@@ -0,0 +1,516 @@
...
+      self.midi_program = 1
+
+  def __str__(self):
+    score_str = "ScorePart: " + self.part_name
+    score_str += ", Channel: " + str(self.midi_channel)
+    score_str += ", Program: " + str(self.midi_program)
+    return score_str
+
+
+class Part:
```
cghawthorne: All of these classes should have a docstring describing what information they contain.

magenta/music/musicxml_parser.py
```diff
@@ -0,0 +1,516 @@
...
+    elif step == "A":
+      pitch_class = 9
+    elif step == "B":
+      pitch_class = 11
+
+    pitch_class = (pitch_class + int(alter)) % 12
+    midi_pitch = (12 + pitch_class) + (int(octave) * 12)
+    return midi_pitch
+
+  def __str__(self):
```
cghawthorne: Is this method outputting a particular format, or just something to make debugging easier to read? Would probably be good to document in the docstring for this method.

magenta/music/musicxml_parser_test.py
```diff
@@ -0,0 +1,170 @@
...
+from collections import defaultdict
+import os.path
+import tempfile
+
+# internal imports
+from musicxml_parser import *
+from musicxml_reader import *
+import tensorflow as tf
+
+# self.flute_scale_filename contains an F-major scale of 8 quarter notes each
```
cghawthorne: This might fit better in the docstring for the class.

magenta/music/musicxml_reader.py
```diff
@@ -0,0 +1,119 @@
...
+        if not musicxml_note.is_rest:
+          note = sequence.notes.add()
+          note.instrument = musicxml_note.midi_channel
+          note.program = musicxml_note.midi_program
+          note.start_time = musicxml_note.time_position
+          note.end_time = musicxml_note.time_position + musicxml_note.seconds
+          note.pitch = musicxml_note.pitch[1] # Index 1 = MIDI pitch number
+          note.velocity = musicxml_note.velocity
+
+  # TODO(@douglaseck): Estimate note type (e.g. quarter note) and populate
```
cghawthorne: Is this TODO still relevant?

iansimon:
I recommend running pylint on your code.  You can even get something close to Google's linter configuration, described in this StackOverflow answer: http://stackoverflow.com/questions/29597618/is-there-a-tool-to-lint-python-based-on-the-google-style-guide

Basically, install pylint, download that googlecl-pylint.rc file, then run:

`pylint --rcfile=/path/to/googlecl-pylint.rc <your files>`

jsawruk:
@iansimon: Thanks. I'm linting the code right now using that pylint configuration. Not done yet, but soon. Then I will address remaining issues raised by @cghawthorne.

magenta/music/musicxml_reader.py
cghawthorne: Yes, that looks correct to me. Feel free to start filling out those fields.

jsawruk:
@iansimon: I am getting a pylint error that I don't know how to resolve. 

Bad option value 'g-import-not-at-top' (bad-option-value)

Please compare with the code here:
https://github.com/tensorflow/magenta/blob/master/magenta/music/midi_io.py#L21

magenta/music/BUILD
```diff
@@ -171,7 +171,7 @@ py_library(
 py_test(
     name = "midi_io_test",
     srcs = ["midi_io_test.py"],
-    data = ["//magenta/testdata"],
+    data = ["//magenta/testdata:testdata"],
```
jsawruk: No idea. Maybe I was having issues initially, but the tests pass now, so it should not have :testdata. Thanks.

magenta/music/BUILD
```diff
@@ -180,6 +180,26 @@ py_test(
 )
 
 py_library(
+    name = "musicxml_reader",
+    srcs = ["musicxml_reader.py", "musicxml_parser.py"],
```
jsawruk: Fixed. Will be in next commit.

magenta/music/musicxml_parser.py
```diff
@@ -0,0 +1,516 @@
...
+# behavior of producing a float when dividing integers
+from __future__ import division
+
+from zipfile import ZipFile
+import xml.etree.ElementTree as ET
+
+# Constants
+
+# Number of MIDI ticks per beat. Required by NoteSequence structure
+TICKS_PER_BEAT = 960
```
jsawruk: Fixed in next commit.

magenta/music/musicxml_parser.py
```diff
@@ -0,0 +1,516 @@
...
+
+# Global variables
+# Default to one division per measure
+current_divisions = 1
+
+# Default to a tempo of 120 beats per minute
+# (This is the default tempo according to the
+# Standard MIDI Files 1.0 Specification)
+current_tempo = 120
```
jsawruk: This is value is mutable. It maintains the state of the currently identified tempo in the MusicXML file. However, there is a place where a default tempo is set if none is found; I have used this constant in that location. It is fixed in the next commit.

magenta/music/musicxml_parser.py
```diff
@@ -0,0 +1,516 @@
...
+
+# Duration of a single beat in seconds
+current_seconds_per_beat = 0.5
```
jsawruk: I don't have a good solution for this at the moment. I know the code isn't threadsafe because I am mutating global variables, but parsing MusicXML files can be tricky. MusicXML files (like most/all? xml files) do not guarantee order of elements, only that the elements exist. For example, let's say there is a tempo marking over beat 1 in the first measure of quarter = 96. Ideally, I'd like to know the tempo before parsing the note on beat one, but MusicXML does not guarantee this. That example is trivial, but it gets more complicated when keeping track of time and you encounter the <forward> and <backup> elements. 

I'd like to keep track of this state somehow, but my only two potential solutions were to pass  around a state object to every class, or to use globals. Since I wanted to get things working quickly, I resorted to globals. I can now refactor to something more elegant and threadsafe. Do you have any alternative solutions to this problem? If not, I'll need to pass around some object that maintains these state variables.

magenta/music/musicxml_parser.py
```diff
@@ -0,0 +1,516 @@
...
+  def __init__(self, xml_score_part = None):
+    self.xml_score_part = xml_score_part
+    self.part_name = ""
+    self.midi_channel = 1
+    self.midi_program = 1
+    if xml_score_part != None:
+      self.parse()
+
+  def parse(self):
+    global current_midi_channel
```
jsawruk: See comment above. Let's keep that discussion in that thread.

magenta/music/musicxml_parser.py
jsawruk: If you're asking about the __str__ function, it was just to make my debugging easier as I was developing. Would you like me to remove the __str__ method altogether, or do you have a proposed format that would be helpful to you?

magenta/testdata/BUILD
```diff
@@ -34,5 +34,9 @@ filegroup(
         "example_is_drum.mid",
         "notesequences.tfrecord",
         "tfrecord_iterator_test.tfrecord",
+        'clarinet_scale.xml',
```
jsawruk: Good idea. This is fixed and will be in the next commit.

magenta/music/musicxml_parser_test.py
jsawruk: Moved to docstring. Will be in the next commit.

magenta/music/musicxml_parser.py
cghawthorne: This is fine, just wanted to make sure I understood the intention.

iansimon:
There seems to be a missing file el_capitan.xml being referred to in your tests.  Did you forget to check this in?

jsawruk:
@iansimon: Thanks, I missed that when I changed where the test files are located.

magenta/music/musicxml_parser.py
```diff
@@ -0,0 +1,628 @@
...
+  @staticmethod
+  def getscore(filename):
+    """Given a MusicXML file, return the score as an xml.etree.ElementTree
+
+    Given a MusicXML file, return the score as an xml.etree.ElementTree
+    If the file is compress (ends in .mxl), uncompress it first
+
+    Args:
+        filename: The path of a MusicXML file
+    """
```
iansimon: You also need a "Returns:" section as part of this docstring.

iansimon:
I fixed many of the lint errors, mainly just ones that required no understanding of the code.  There are still several that remain, largely missing/malformed docstring sections.  There's also a CURRENT_DIVISIONS variable in musicxml_parser.py that is commented out in some places but not others.

jsawruk:
@iansimon: Thanks for fixing the lint errors. I have been extremely busy at work and was hoping to work more on this code at the Classical Music Hack Day at MIT this coming weekend.

jsawruk:
I believe this latest commit addresses all issues raised so far. I have implemented the note.numerator and note.denominator support (@douglaseck's TODO)

magenta/music/musicxml_parser.py
```diff
@@ -0,0 +1,678 @@
...
+    for child in self.xml_measure:
+      if child.tag == "attributes":
+        self.parseattributes(child)
+      elif child.tag == "backup":
+        self.parsebackup(child)
+      elif child.tag == "direction":
+        self.parsedirection(child)
+      elif child.tag == "forward":
+        self.parseforward(child)
+      elif child.tag == "note":
```
cghawthorne: should there also be an else condition that throws an exception for an unrecognized tag?

magenta/music/musicxml_parser.py
```diff
@@ -0,0 +1,678 @@
...
+
+    for child in xml_attributes:
+      if child.tag == "divisions":
+        #CURRENT_DIVISIONS = int(child.text)
+        self.state.divisions = int(child.text)
+      elif child.tag == "key":
+        self.key_signature = KeySignature(self.state, child)
+      elif child.tag == "time":
+        self.time_signature = TimeSignature(self.state, child)
+      elif child.tag == "transpose":
```
cghawthorne: add else branch?

magenta/music/musicxml_parser.py
```diff
@@ -0,0 +1,678 @@
...
+        self.parsepitch(child)
+      elif child.tag == "rest":
+        self.is_rest = True
+      elif child.tag == "voice":
+        self.voice = int(child.text)
+      elif child.tag == "dot":
+        self.note_duration.dots += 1
+      elif child.tag == "type":
+        self.note_duration.type = child.text
+      elif child.tag == "time-modification":
```
cghawthorne: add else branch?

magenta/music/musicxml_parser.py
```diff
@@ -0,0 +1,678 @@
...
+
+    # Visual pitch representation
+    alter_string = ""
+    if alter == "-2":
+      alter_string = "bb"
+    elif alter == "-1":
+      alter_string = "b"
+    elif alter == "1":
+      alter_string = "#"
+    elif alter == "2":
```
cghawthorne: add else branch?

magenta/music/musicxml_parser.py
```diff
@@ -0,0 +1,678 @@
...
+      pitch_class = 2
+    elif step == "E":
+      pitch_class = 4
+    elif step == "F":
+      pitch_class = 5
+    elif step == "G":
+      pitch_class = 7
+    elif step == "A":
+      pitch_class = 9
+    elif step == "B":
```
cghawthorne: add else branch?

magenta/music/musicxml_parser.py
```diff
@@ -0,0 +1,678 @@
...
+
+  def converttypetoratio(self):
+    """Convert the MusicXML note-type-value to a Python Fraction
+    Examples:
+    - whole = 1/1
+    - half = 1/2
+    - quarter = 1/4
+    - 32nd = 1/32
+    """
+    typeratiomap = {"maxima": Fraction(8, 1), "long": Fraction(4, 1),
```
cghawthorne: this should probably be a class-level constant

magenta/music/musicxml_parser_test.py
```diff
@@ -0,0 +1,190 @@
...
+
+from collections import defaultdict
+import os.path
+
+# internal imports
+import tensorflow as tf
+from magenta.music.musicxml_parser import MusicXMLDocument
+from magenta.music.musicxml_reader import musicxml_to_sequence_proto
+
+class MusicXMLParserTest(tf.test.TestCase):
```
cghawthorne: It looks like this test just verifies that the parsed musicxml data is consistent with the generated note sequence proto. There should also be some tests to verify that the data is parsed correctly. For example, you could verify all the properties you have described here in text (e.g., verify that the expected F-major scale is parsed from clarinet_scale).

magenta/music/musicxml_reader.py
```diff
@@ -0,0 +1,116 @@
...
+# limitations under the License.
+"""MusicXML import.
+
+Input wrappers for converting MusicXML into tensorflow.magenta.NoteSequence.
+"""
+
+# internal imports
+from magenta.music.musicxml_parser import MusicXMLDocument
+from magenta.protobuf import music_pb2
+# pylint: enable=g-import-not-at-top
```
cghawthorne: why is this needed?

magenta/music/musicxml_reader.py
```diff
@@ -0,0 +1,116 @@
...
+
+  # Populate tempo changes.
+  musicxml_tempos = musicxml_document.gettempos()
+  for musicxml_tempo in musicxml_tempos:
+    tempo = sequence.tempos.add()
+    tempo.time = musicxml_tempo.time_position
+    tempo.qpm = musicxml_tempo.qpm
+
+  # Populate notes from each MusicXML part across all voices
+  # Unlike MIDI import, notes are not sorted
```
cghawthorne: any reason for not sorting them?

magenta/music/musicxml_parser.py
```diff
@@ -0,0 +1,678 @@
...
+    self.seconds = (self.midi_ticks / constants.STANDARD_PPQ)
+    self.seconds *= self.state.seconds_per_beat
+
+    self.time_position = self.state.time_position
+
+    if note.is_in_chord:
+      # If this is a chord, set the time position to the time position
+      # of the previous note (i.e. all the notes in the chord will have
+      # the same time position)
+      self.time_position = self.state.previous_note.time_position
```
cghawthorne: When I tried to run convert_musicxml_dir_to_note_sequences on the test xml files, I got an error caused by this line:

```
...
  File ".../magenta/bazel-bin/magenta/scripts/convert_musicxml_dir_to_note_sequences.runfiles/__main__/magenta/music/musicxml_parser.py", line 528, in parseduration
    self.time_position = self.state.previous_note.time_position
AttributeError: 'Note' object has no attribute 'time_position'
```

Do you know why the unit tests didn't catch this?

magenta/music/musicxml_parser.py
```diff
@@ -0,0 +1,678 @@
+# Licensed under the Apache License, Version 2.0 (the "License");
```
jsawruk: Done. Available in next commit.

magenta/music/musicxml_parser.py
```diff
@@ -0,0 +1,678 @@
...
+import xml.etree.ElementTree as ET
+from fractions import Fraction
+from zipfile import ZipFile
+from magenta.music import constants
+
+
+class MusicXMLParserState(object):
+  """Maintains internal state of the MusicXML parser"""
+  def __init__(self):
+    # Default to one division per measure
```
jsawruk: Divisions is difficult to explain. I added verbiage from the MusicXML documentation that I thought was the easiest to understand. I couldn't fit the entire URL as a citation, but you can find more info here: 
http://usermanuals.musicxml.com/MusicXML/MusicXML.htm#EL-MusicXML-divisions.htm

This will additional comment will be in the next commit.

magenta/music/musicxml_parser.py
```diff
@@ -0,0 +1,678 @@
...
+    self.divisions = 1
+
+    # Default to a tempo of 120 beats per minute
+    # (This is the default tempo according to the
+    # Standard MIDI Files 1.0 Specification)
+    self.tempo = 120
```
jsawruk: Ok, I've changed it and the corresponding documentation. I originally called it tempo because in MusicXML it is called "tempo". Since Magenta calls it qpm, I also updated the comment to explain that the two are equivalent.

magenta/music/musicxml_parser.py
```diff
@@ -0,0 +1,678 @@
...
+
+    # Duration of a single beat in seconds
+    self.seconds_per_beat = 0.5
```
jsawruk: Yes, since everything is now in quarter, this variable should be renamed as well. Commented updated to reflect this.

magenta/music/musicxml_parser.py
```diff
@@ -0,0 +1,678 @@
...
+
+    # Running total of time for the current event.
+    # Resets to 0 on every part. Affected by <forward> and <backup> elements
```
jsawruk: The unit is indeed seconds, so I updated the comment.

magenta/music/musicxml_parser.py
```diff
@@ -0,0 +1,678 @@
...
+    # Default to a MIDI velocity of 64 (mf)
+    self.velocity = 64
+
+    # Default MIDI program (1 = grand piano)
+    self.midi_program = 1
+
+    # Current MIDI channel (usually equal to the part number)
+    self.midi_channel = 1
+
+    # Keep track of previous note to get chord timing correct
```
jsawruk: Updated this comment. It stores an instance of the Note class.

magenta/music/musicxml_parser.py
```diff
@@ -0,0 +1,678 @@
...
+    self.previous_note = None
+
+    # Keep track of current transposition level
```
jsawruk: Yes, it can be positive or negative semitones, so I updated the comment. Note that it does not support quarter tones or tuning systems other than equal temperament.

magenta/music/musicxml_parser.py
```diff
@@ -0,0 +1,678 @@
...
+  """
+  def __init__(self, filename):
+    self.filename = filename
+    self.score = self.getscore(filename)
+    self.parts = []
+    self.score_parts = []
+    self.midi_resolution = constants.STANDARD_PPQ
+    self.state = MusicXMLParserState()
+    # Total time in seconds
+    self.total_time = 0
```
jsawruk: Updated. Will be in next commit.

magenta/music/musicxml_parser.py
```diff
@@ -0,0 +1,678 @@
...
+    self.parse()
+
+  @staticmethod
+  def getscore(filename):
```
jsawruk: Updated in commit 4288357.

magenta/music/musicxml_parser.py
```diff
@@ -0,0 +1,678 @@
...
+    """
+    if filename.endswith('.mxl'):
+      # Compressed MXL file. Uncompress in memory.
+      filename = ZipFile(filename)
+
+      # Ignore files in the META-INF directory.
+      # Only retrieve the first file
+      namelist = filename.namelist()
+      files = [name for name in namelist if not name.startswith('META-INF/')]
+      compressed_file_name = files[0]
```
jsawruk: I'm holding off on this one until I can get an MXL file with multiple rootfiles and add it to the testing to ensure we can handle this case. I didn't see a way to export directly to this file type from the notation programs I use at work, so I will probably have to create a simple example by hand.

magenta/music/musicxml_parser.py
```diff
@@ -0,0 +1,678 @@
...
+      score = ET.fromstring(filename.read(compressed_file_name))
+    else:
+      # Uncompressed XML file.
+      tree = ET.parse(filename)
+      score = tree.getroot()
+
+    return score
+
+  def parse(self):
```
jsawruk: I assume you mean prepend with double underscores. This is done and available in commit 62da424

magenta/music/musicxml_parser.py
```diff
@@ -0,0 +1,678 @@
...
+    xml_part_list = self.score.find("part-list")
+    for element in xml_part_list:
+      if element.tag == "score-part":
+        score_part = ScorePart(self.state, element)
+        self.score_parts.append(score_part)
+
+    # Parse parts
+    xml_parts = self.score.findall("part")
+    score_part_index = 0
+    for child in xml_parts:
```
jsawruk: Yes. Thanks for the suggestion. This will be available in the next commit.

magenta/music/musicxml_parser.py
```diff
@@ -0,0 +1,678 @@
...
+      key_signatures.append(key_signature)
+
+    return key_signatures
+
+  def gettempos(self):
+    """Return a list of all tempos in this score.
+    If no tempos are found, create a default tempo of 120 qpm.
+    """
+    tempos = []
+    #for part in self.parts:
```
jsawruk: Not sure what you want deleted here. I removed the comment, but I kept the initialization of the tempos[] array.

magenta/music/musicxml_parser.py
```diff
@@ -0,0 +1,678 @@
...
+    if xml_score_part != None:
+      self.parse()
+
+  def parse(self):
+    """Parse the <score-part> element to an in-memory representation"""
+    if self.xml_score_part.find("part-name") != None:
+      self.part_name = self.xml_score_part.find("part-name").text
+
+    xml_midi_instrument = self.xml_score_part.find("midi-instrument")
+    if xml_midi_instrument != None and \
```
jsawruk: Done. Available in commit a462555.

magenta/music/musicxml_parser.py
```diff
@@ -0,0 +1,678 @@
...
+    self.measures = []
+    self.transposes = False
+    self.state = state
+    self.parse()
+
+  def parse(self):
+    """Parse the <part> element"""
+
+    # Reset the time position when parsing each part
+    #CURRENT_TIME_POSITION = 0
```
jsawruk: Yes, forgot to remove this when refactoring. Removed in next commit.

magenta/music/musicxml_parser.py
```diff
@@ -0,0 +1,678 @@
...
+    self.state.time_position = 0
+    self.state.midi_channel = self.score_part.midi_channel
+    self.state.midi_program = self.score_part.midi_program
+    self.state.transpose = 0
+
+    xml_measures = self.xml_part.findall("measure")
+    for child in xml_measures:
+      measure = Measure(child, self, self.state)
```
jsawruk: Not sure I follow here. Are you saying that Measure should not mutate the state.transpose property? Also, the Measure class already has a transpose property.

The problem is that a single part may transpose within a given measure. For example, a musician may initially play a clarinet part (Bb), but then change instruments to alto sax (Eb). This transposition persists for subsequent measures until another transpose attribute is found. 

Example: In Part 1, Measure 1 defines transposition as Bb (-2). Measure 2 does not have a transpose attribute, so it is still -2. I need to persist this state somewhere, so I put it on the state object for now. Other suggestions greatly appreciated.

magenta/music/musicxml_parser.py
```diff
@@ -0,0 +1,678 @@
...
+class Measure(object):
+  """Internal represention of the MusicXML <measure> element"""
+  def __init__(self, xml_measure, part, state):
+    self.xml_measure = xml_measure
+    self.notes = []
+    self.tempos = []
+    self.time_signature = None
+    self.key_signature = KeySignature(state) # Default to C major
+    self.current_ticks = 0      # Cumulative tick counter for this measure
+    self.transpose = 0          # Number of semitones to transpose notes
```
jsawruk: I'm honestly not sure. I put this property here because the <measure> element is where the transpose property is located in MusicXML. Whether this needs to be part of the Measure class or not is contingent upon the question you asked above.

magenta/music/musicxml_parser.py
jsawruk: A `<measure>` element can contain tags that I am not explicitly parsing here because I didn't feel they were relevant to Magenta (e.g. `<barline>`). Would you like these tags parsed but no operation when encountered, and then an exception for a truly unrecognized tag?

Also, since there is an XSD/DTD for MusicXML, it might be possible to instead just validate against those during the initial parsing. Thoughts?

magenta/music/musicxml_parser.py
```diff
@@ -0,0 +1,678 @@
...
+        self.notes.append(note)
+        # Keep track of current note as previous note for chord timings
+        self.state.previous_note = note
+
+  def parseattributes(self, xml_attributes):
+    """Parse the MusicXML <attributes> element"""
+
+    for child in xml_attributes:
+      if child.tag == "divisions":
+        #CURRENT_DIVISIONS = int(child.text)
```
jsawruk: Deleted. Accidentally left over from my refactoring. Will be in next commit.

magenta/music/musicxml_parser.py
jsawruk: Like `<measure>`, `<attributes>` has several different elements, and I am only parsing the ones that I felt were relevant to Magenta. Please see comments above regarding `<measure>`.

magenta/music/musicxml_parser.py
```diff
@@ -0,0 +1,678 @@
...
+
+    self.midi_channel = self.state.midi_channel
+    self.midi_program = self.state.midi_program
+    self.velocity = self.state.velocity
+
+    for child in self.xml_note:
+      if child.tag == "chord":
+        self.is_in_chord = True
+      elif child.tag == "duration":
+        self.note_duration.parseduration(self, child.text)
```
jsawruk: Good catch. Available in next commit.

magenta/music/musicxml_parser.py
jsawruk: See comments above about `<measure>` and `<attributes>`.

magenta/music/musicxml_parser.py
jsawruk: The `<alter>` tag can support decimal semitones, but I chose to not support this because MIDI doesn't and the model currently used in Magenta uses MIDI number. Perhaps a better option would be to cast the string to a decimal, then round to an int. The else branch would then just be an empty string (no accidental)?

magenta/music/musicxml_parser.py
jsawruk: There is no else branch, as all the possible valid cases are covered. 
http://usermanuals.musicxml.com/MusicXML/MusicXML.htm#EL-MusicXML-step.htm

The only thing an else branch could do would be to throw an exception. Would you like me to add that in this case? What kind of exception would it be?

magenta/music/musicxml_parser_test.py
jsawruk: I initially created the test cases based on what the MIDI test cases did at that time and added one for the compressed MusicXML file. What other test cases do you think would be useful?

magenta/music/musicxml_parser_test.py
```diff
@@ -0,0 +1,190 @@
...
+                               * musicxml_note.note_duration.durationfloat(), \
+                               delta=1)
+
+  def checkmusicxmltosequence(self, filename):
+    """Test the translation from MusicXML to Sequence proto."""
+    source_musicxml = MusicXMLDocument(filename)
+    sequence_proto = musicxml_to_sequence_proto(source_musicxml)
+    self.checkmusicxmlandsequence(source_musicxml, sequence_proto)
+
+  #def testsimplemusicxmltosequence(self):
```
jsawruk: Because I was debugging the rhythm test and forgot to uncomment them :). Fixed in commit 4ebfb5a

magenta/music/musicxml_reader.py
jsawruk: I didn't understand the rationale for sorting the MIDI notes, so I didn't sort the notes from the MusicXML. What purpose does sorting serve?

magenta/music/musicxml_parser.py
jsawruk: Not sure why the unit tests didn't catch this, but this code has been corrected.

magenta/music/musicxml_reader.py
cghawthorne: On second thought, I believe you are correct. We should not rely on these notes being sorted. I've created #315 to make our documentation more clear.

magenta/music/musicxml_reader.py
cghawthorne: I'm wondering why the pylint comment is needed. I don't see any imports out of place, so there shouldn't be any need to disable that linter check.

magenta/music/musicxml_parser.py
cghawthorne: What I'm objecting to is passing `self` into the constructor of `Measure` because that's generally not a good programming practice.

It looks like the only reason to pass `self` in here is so that the `transposes` property of this object can be modified based on whether `Measure` transposes or not. It seems like that information could be communicated either through a property of `Measure`, or maybe just modifying the state object would be sufficient.

magenta/music/musicxml_parser.py
cghawthorne: Ah, I see. Could you add a comment explaining that, just so it's clear that there are other valid tags and we're just choosing to ignore them. Maybe something like:

```
else:
  # Ignore other tag types because they are not relevant to Magenta.
  pass
```

magenta/music/musicxml_parser.py
cghawthorne: Ditto for adding a comment

magenta/music/musicxml_parser.py
cghawthorne: For this one, maybe add an else branch that logs a message at the `debug` level, along with a comment that we're currently ignoring decimal semitones. We've thought about adding support for that sort of thing in the (likely distant) future, so it would be good to have an obvious place to go back and add it.

magenta/music/musicxml_parser.py
cghawthorne: Yeah, I think it would be good to throw an exception here because reaching the else branch in this case would indicate either a corrupt file or a subtle bug in the parser. I would prefer a custom exception for this class. Here's an example where we've used a custom exception before:https://github.com/tensorflow/magenta/blob/master/magenta/music/sequence_generator_bundle.py

magenta/music/musicxml_parser_test.py
cghawthorne: I think it would be good to verify the things described below. For example:
- flute_scale should have an F-major scale of 8 quarter notes. You could verify that the notes and durations are parsed correctly.
- clarinet_scale should definitely verify that the note values are parsed and transposed correctly.
- band_score could verify a few chords, time signature changes, tempo changes, etc.

Our MIDI test cases are admittedly sparse. For a better example, take a look at our QuantizedSequence tests, which construct a NoteSequence and an expected QuantizedSequence that should result: https://github.com/tensorflow/magenta/blob/master/magenta/music/sequences_lib_test.py#L128

magenta/music/musicxml_parser.py
```diff
@@ -0,0 +1,692 @@
...
+
+      # Ignore files in the META-INF directory.
+      # Only retrieve the first file
+      namelist = filename.namelist()
+      files = [name for name in namelist if not name.startswith('META-INF/')]
+      compressed_file_name = files[0]
+      score = ET.fromstring(filename.read(compressed_file_name))
+    else:
+      # Uncompressed XML file.
+      tree = ET.parse(filename)
```
cghawthorne: The call to `parse` should be surrounded by a try/except because it can raise `xml.etree.ElementTree.ParseError`. I noticed this while trying to parse the testdata directory. The `convert_musicxml_dir_to_note_sequences` script is able to catch `MusicXMLConversionError` exceptions and continue, but the `ParseError` exception caused the script to stop because it wasn't caught by anything.

cghawthorne:
I was just doing some manual testing and noticed some potential issues with parsing. Some of these should probably be added as unit tests.

- flute_scale.mxl has a key_signature of C added at time -1.0

- el_capitan.xml has a key signature of C at time -1.0 too, and it also has some duplicate key signatures. I'm wondering if the duplicate key signatures are due to the transposing parts? Should everything be transposed to 1 key signature before writing? Here's the current key signature output:
```
key_signatures {
  time: 0.0
  key: B_FLAT
}
key_signatures {
  time: -1.0
  key: C
}
key_signatures {
  time: 81.0
  key: E_FLAT
}
key_signatures {
  time: 81.0
  key: F
}
key_signatures {
  time: 0.0
  key: D_FLAT
}
key_signatures {
  time: 81.0
  key: C
}
key_signatures {
  time: 81.0
  key: B_FLAT
}
```

- Not a bug, but worth verifying in a unit test: I compared how our music21-based parser and your parser handled clarinet_scale, and it turns out the music21-based parser got it wrong (it didn't transpose) and yours got it right! Thanks for figuring out all the tricky transposition stuff.

For testing this, I just used some code that looked like this:
```
import magenta.music as mm
mxml = mm.note_sequence_io.note_sequence_record_iterator('/tmp/jmusicxml.tfrecord')
mxml.next()
```
Note that when protobufs are printed out, they don't include default values or 0s. So if a time_signature happens at time 0, it will print out without a time value. Similarly, because the key signature enum has C as 0, key signatures with a value of C will print out without a key.

magenta/music/musicxml_parser.py
jsawruk: For some reason, I thought it would be useful to indicate that the Part transposes, and Measure was mutating that flag on the Part since transposition is identified at the measure level in MusicXML (see my comments above for an example use case).

I am not using this attribute on the Part object at all, so it looks like it was just old code that needs to be cleaned up. I have removed Part.transposes and am now no longer passing the Part into the Measure constructor. Fixed in commit c581826.

magenta/music/musicxml_parser.py
jsawruk: It appears that this code is no longer necessary; it has indeed been replaced by self.state.transpose. Removed in commit 5c3b6a0.

magenta/music/musicxml_parser.py
jsawruk: Added per your suggestion below:
```
else:
  # Ignore other tag types because they are not relevant to Magenta.
  pass
```
Added in next commit.

magenta/music/musicxml_parser.py
jsawruk: Added per your suggestion.
Added in next commit.

magenta/music/musicxml_parser.py
jsawruk: Added per your suggestion in next commit.

magenta/music/musicxml_parser.py
jsawruk: Exception added in next commit.

magenta/music/musicxml_reader.py
jsawruk: Not sure how that got there either. Removed in next commit.

magenta/music/musicxml_parser.py
jsawruk: I added a try/except around both ET.fromstring and ET.parse in the event of ET.ParseError. This has been added in commit 02fc66f.

magenta/music/musicxml_parser.py
```diff
@@ -110,12 +113,26 @@ def get_score(filename):
...
       compressed_file_name = files[0]
-      score = ET.fromstring(filename.read(compressed_file_name))
+      try:
+        score = ET.fromstring(filename.read(compressed_file_name))
+      except ET.ParseError as parse_error:
+        line_number, column = parse_error.position
+        line = next(IT.islice(io.BytesIO(content), line_number))
+        caret = '{:=>{}}'.format('^', column)
+        parse_error.msg = '{}\n{}\n{}'.format(parse_error, line, caret)
+        print(parse_error.msg)
```
cghawthorne: Rather than printing the parse error, it would be better to raise a custom exception that contains the parse error, so the caller can decide what to do with the error (e.g., skip the file silently, print out an error, add it to an error log, etc.).

Wrapping the ET.ParseError exception in a custom exception is important because callers into this class shouldn't need to know that this class uses the ElementTree class. They should just need to know how to handle exceptions that this class might raise.

magenta/music/musicxml_parser_test.py
jsawruk: Added unit tests for transposition in commit 5afa30c.

magenta/music/musicxml_parser.py
jsawruk: Updated with new MusicXMLParseException class in commit 5afa30c.

magenta/music/musicxml_parser.py
jsawruk: Changed in commit fa38cb8.

cghawthorne:
Hey Jeremy, I think this is looking pretty good! We might need some followup changes, but I think it's in a good shape to check in now!

PR #226, opened 2016-09-28
update readme files with Docker path notes

elburz:
- added notes for windows and mac users on main readme re: shared folder paths
- added note to main readme to verify the output_dir flag matches their defined shared folders from the docker run command
- added notes on basic_rnn, lookback_rnn, and attention_rnn readme's noting that the .mag files are already present on the Docker container

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,8 +23,10 @@ ready to run. It will also map port 6006 of the host machine to the container so
 you can view TensorBoard servers that run within the container.
 
 This also maps the directory ```/tmp/magenta``` on the host machine to
-```/magenta-data``` within the Docker session. **WARNING**: only data saved in
-```/magenta-data``` will persist across Docker sessions.
+```/magenta-data``` within the Docker session. Windows users can change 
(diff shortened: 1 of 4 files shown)


elburz:
I signed it!

README.md
```diff
@@ -23,8 +23,10 @@ ready to run. It will also map port 6006 of the host machine to the container so
 you can view TensorBoard servers that run within the container.
 
 This also maps the directory ```/tmp/magenta``` on the host machine to
-```/magenta-data``` within the Docker session. **WARNING**: only data saved in
-```/magenta-data``` will persist across Docker sessions.
+```/magenta-data``` within the Docker session. Windows users can change 
+```/tmp/magenta``` to a path such as ```C:/magenta```, and Mac users can
```
cghawthorne: nit: change to "Mac and Linux users use a path relative to their home folder"

elburz:
Added the above change @cghawthorne thanks for the note.

PR #241, opened 2016-10-04
play synthesized note sequences in Jupyter notebooks

iansimon:


magenta/music/midi_synth.py
```diff
@@ -0,0 +1,55 @@
...
+"""MIDI audio synthesis."""
+
+# internal imports
+import numpy as np
+
+from magenta.music import midi_io
+
+
+def synthesize(sequence, sample_rate, wave=np.sin):
+  """Synthesizes audio from a music_pb2.NoteSequence using a waveform.
```
adarob: We should probably mention that the quality will be lower than using fluidsynth with a good soundfont.

iansimon: Added text mentioning such.

PR #265, opened 2016-10-07
Fix timing

adarob:


adarob:
@jesseengel, I accidentally merged this prematurely. Please review and we will revert if needed.

magenta/models/shared/melody_rnn_sequence_generator.py
```diff
@@ -195,7 +196,7 @@ def generate_melody(self, num_steps, primer_melody):
     softmax = self._session.graph.get_collection('softmax')[0]
 
     final_state_ = None
-    for i in range(num_steps - len(melody)):
+    for i in range(num_steps - (len(melody) + melody.start_step)):
```
jesseengel: Aha, so it was an offset problem. Good catch.

PR #266, opened 2016-10-08
added installation script

jesseengel:
Following up on the hackathon, I made an install shell script that completely installs magenta.

The script does not use homebrew, and uses rtmidi instead of portmidi, which we found to work better and have native support on OSX during the hackathon. This requires switching the mido backend in magenta code, which is probably a good idea anyways. 

It detects if anaconda is already installed. If so, it creates a new magenta environment and downloads all dependencies. If not, it first downloads and installs miniconda, and then installs an environment. 

python-rtmidi seems to work out of the box on OSX, but needs a few libs installed on linux. The script installs these.

I've tested this on my macbook air and linux workstation, but it will likely need a little tweaking to work in all possible environments.


install_magenta.sh
```diff
@@ -0,0 +1,84 @@
+#!/bin/bash
+
+# Check which operating system
+if [ "$(uname)" == "Darwin" ]; then
```
cghawthorne: [[ is preferred over [
https://google.github.io/styleguide/shell.xml#Test,_%5B_and_%5B%5B

install_magenta.sh
```diff
@@ -0,0 +1,84 @@
...
+    echo "Mac OS Detected"
+    export OS="MAC"
```
cghawthorne: since these don't need to be exported to the environment, 'readonly' might be a better choice.

https://google.github.io/styleguide/shell.xml?showone=Constants_and_Environment_Variable_Names#Constants_and_Environment_Variable_Names

install_magenta.sh
```diff
@@ -0,0 +1,84 @@
...
+
+
+# Check if anaconda already installed
+if [ ! $(which conda) ]; then
+    echo ""
+    echo "==========================================="
+    echo "anaconda not detected, installing miniconda"
+    echo "==========================================="
+    echo ""
+    curl https://repo.continuum.io/miniconda/$MINICONDA_SCRIPT > $/tmp/$MINICONDA_SCRIPT
```
cghawthorne: should quote these strings: https://google.github.io/styleguide/shell.xml?showone=Constants_and_Environment_Variable_Names#Quoting

install_magenta.sh
```diff
@@ -0,0 +1,84 @@
...
+echo ""
+
+conda create -n magenta python=2.7
+source activate magenta
+
+# Install tensorflow
+pip install --ignore-installed --upgrade $TF_BINARY_URL
+
+# Install other dependencies
+pip install jupyter magenta
```
cghawthorne: I'm not completely sure, but I think if you include jupyter on the 'conda create' command or use 'conda install' , it can use the cached jupyter package, which the user might already have if they have installed the full anaconda environment before.

install_magenta.sh
```diff
@@ -0,0 +1,84 @@
...
+
+# Install rtmidi for realtime midi IO
+if [ $OS == "LINUX" ]; then
+    echo ""
+    echo "============================================"
+    echo "Installing rtmidi Linux library dependencies"
+    echo "Sudo privileges required"
```
cghawthorne: nit: Sudo shouldn't be capitalized.

install_magenta.sh
```diff
@@ -0,0 +1,84 @@
...
+    echo "============================================"
+    echo ""
+    sudo apt-get install sudo apt-get install build-essential
```
cghawthorne: would it be better to just list these all as one command?

install_magenta.sh
```diff
@@ -0,0 +1,84 @@
+#!/bin/bash
+
```
cghawthorne: I'd suggest using 'set -e' at the top because we probably don't want to continue if any step fails.

install_magenta.sh
```diff
@@ -0,0 +1,84 @@
...
+# Check which operating system
+if [ "$(uname)" == "Darwin" ]; then
+    echo "Mac OS Detected"
+    export OS="MAC"
+    export MINICONDA_SCRIPT=Miniconda2-latest-MacOSX-x86_64.sh
+    # Mac OS X, CPU only, Python 2.7:
+    export TF_BINARY_URL=https://storage.googleapis.com/tensorflow/mac/cpu/tensorflow-0.11.0rc0-py2-none-any.whl
```
cghawthorne: We'll want to add a note to our release process to make sure we update these URLs to match our setup.py version requirements.

install_magenta.sh
```diff
@@ -0,0 +1,84 @@
+#!/bin/bash
```
cghawthorne: add Google copyright header

install_magenta.sh
```diff
@@ -0,0 +1,84 @@
...
+    export MINICONDA_SCRIPT=Miniconda2-latest-MacOSX-x86_64.sh
+    # Mac OS X, CPU only, Python 2.7:
+    export TF_BINARY_URL=https://storage.googleapis.com/tensorflow/mac/cpu/tensorflow-0.11.0rc0-py2-none-any.whl
+elif [ "$(uname)" == "Linux" ]; then
+    echo "Linux OS Detected"
+    export OS="LINUX"
+    export MINICONDA_SCRIPT=Miniconda2-latest-Linux-x86_64.sh
+    # Ubuntu/Linux 64-bit, CPU only, Python 2.7
+    export TF_BINARY_URL=https://storage.googleapis.com/tensorflow/linux/cpu/tensorflow-0.11.0rc0-cp27-none-linux_x86_64.whl
+fi
```
cghawthorne: should have an else clause that prints an error if the user is running an OS we don't recognize that encourages them to follow the manual installation instructions.

install_magenta.sh
```diff
@@ -0,0 +1,84 @@
+#!/bin/bash
+
+# Check which operating system
+if [ "$(uname)" == "Darwin" ]; then
+    echo "Mac OS Detected"
+    export OS="MAC"
+    export MINICONDA_SCRIPT=Miniconda2-latest-MacOSX-x86_64.sh
+    # Mac OS X, CPU only, Python 2.7:
+    export TF_BINARY_URL=https://storage.googleapis.com/tensorflow/mac/cpu/tensorflow-0.11.0rc0-py2-none-any.whl
+elif [ "$(uname)" == "Linux" ]; then
```
cghawthorne: I'm not sure the best way to do this, but we probably only want to proceed if they're running a Debian variant since we expect to call apt-get later in the script. I think you could check for the presence of /etc/debian_version, or maybe just see if apt-get is available on the path.

install_magenta.sh
jesseengel: I think it uses a cached version if it meets the requirements either way

install_magenta.sh
jesseengel: done

install_magenta.sh
jesseengel: done

install_magenta.sh
jesseengel: done

install_magenta.sh
```diff
@@ -0,0 +1,84 @@
...
+    echo "============================================"
+    echo "Installing rtmidi Linux library dependencies"
+    echo "Sudo privileges required"
+    echo "============================================"
+    echo ""
+    sudo apt-get install sudo apt-get install build-essential
+    sudo apt-get install libasound2-dev 
+    sudo apt-get install libjack-dev
+fi
+pip install python-rtmidi
```
adarob: i'm pretty sure you'll need the '--pre' flag for this to work in all cases

install_magenta.sh
jesseengel: done

install_magenta.sh
jesseengel: done

install_magenta.sh
jesseengel: done

install_magenta.sh
jesseengel: done

install_magenta.sh
jesseengel: done

install_magenta.sh
jesseengel: done

install_magenta.sh
jesseengel: done

install_magenta.sh
jesseengel: done

install_magenta.sh
jesseengel: done

magenta/tools/build.sh
```diff
@@ -0,0 +1,35 @@
...
+
+# Exit on error
+set -e
+
+
+# Get script directory
+readonly DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+
+# build the binary to make pip package
+bazel build //magenta/tools/pip:build_pip_package
```
cghawthorne: Should probably cd to $DIR before running this command, in case the user was outside the magenta tree when they ran the script.

magenta/tools/build.sh
```diff
@@ -0,0 +1,35 @@
+#!/bin/bash
+#
+# Builds the pip package, and then installs it.
```
cghawthorne: I think we typically have the copyright header first and follow it with any descriptive comments

magenta/tools/install.sh
```diff
@@ -0,0 +1,117 @@
+#!/bin/bash
+#
+# An install script for magenta (https://github.com/tensorflow/magenta).
```
cghawthorne: ditto about comment order

magenta/tools/install.sh
```diff
@@ -0,0 +1,117 @@
...
+    echo 'Linux OS Detected'
+    readonly OS='LINUX'
+    readonly MINICONDA_SCRIPT='Miniconda2-latest-Linux-x86_64.sh'
+    # Ubuntu/Linux 64-bit, CPU only, Python 2.7
+    readonly TF_BINARY_URL='https://storage.googleapis.com/tensorflow/linux/cpu/tensorflow-0.11.0rc0-cp27-none-linux_x86_64.whl'
+else
+    err 'Detected neither OSX or Linux Operating System'
+fi
+
+echo ${MINICONDA_SCRIPT}
```
cghawthorne: was this for debugging? if not,  should we also output the TF_BINARY_URL?

PR #273, opened 2016-10-13
Maintain Metronome

hanzorama:
-for the interactions, it is desirable to differentiate between
starting/stopping the metronome and mute/unmute. The latter will allow
for external sync during playback and capture by maintaining the
metronome reference in the background even though inaudible.
-Will also allow for easier implementation of count-offs
-update: switch mute/unmute to a state change. Will add velocity update separately

### DIFF
--- a/magenta/interfaces/midi/midi_hub.py
+++ b/magenta/interfaces/midi/midi_hub.py
@@ -166,6 +166,8 @@ class Metronome(threading.Thread):
     self._duration = duration
     # A signal for when to stop the metronome.
     self._stop_time = stop_time
+    # A Boolean specifying mute state of the metronome.
+    self._mute = False
     super(Metronome, self).__init__()
 
   def run(self):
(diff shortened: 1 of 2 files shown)


jesseengel:
This feature doesn't hurt, but fwiw, I think syncing with pulse messages is real messy, best handled by protocols such as Link (https://github.com/Ableton/link). For us, it seems like the easiest option is to have start/stop events controlled by events sent from the host software rather than trying to sync a host to some metronome of our own.

hanzorama:
I agree that Link is a nice way to talk between devices, but it gets a bit
tricky because it is doing some time conversion of its own between wall
time and sample based time. One of my main concerns is how tempo changes
are currently handled internally within magenta (which is "not at all")
independent of any sync mechanism. How the tempo is passed to and from the
hub based on the wall clock was decided under the assumption that magenta
would always be the master and not have update-able tempo during capture or
playback. So Link might be able to inform that there is a tempo change
externally, but how it would be handled is the question. And while
midi-clock messages to slaves are less resolute and more jitter prone, it
communicates well with any MIDI capable software and hardware devices. So I
am totally into using sync tools like Link but think there are some other
fundamental timing considerations to adjust to really make it useful.

adarob:
Is this change essentially equivalent to stopping and restarting the metronome with the same start_time? That was how I was handling this for now.

hanzorama:
The metronome remains constant without joining the thread, just the clicks
are audible or not. I'll add midi clock support to that later tonight which
will require a continuous metronome.

also re:Link...I also think Link only works for Ableton to send clock
information. When receiving I think it has to be in external mode so then
it relies on midi clock or midi timecode anyway. But I have to revisit that
to be sure

adarob:
I don't think this will be necessary any longer.

PR #275, opened 2016-10-17
use beam search to generate melodies

iansimon:
Modifies the melody generation to use beam search instead of taking "greedy" steps.  This version of beam search uses three parameters:

_beam_size_: The number of melodies to include in the "beam"; the current active set.  When using a pre-trained bundle, this parameter must match the bundle's generation batch size.

_branch_factor_: The number of new melodies to generate per active melody in each beam search iteration.  All but the _beam_size_ melodies with highest likelihood will then be discarded.

_steps_per_iteration_: The number of melody steps to generate per beam search iteration.

When all three of these parameters have value 1, the generation matches the old "greedy" behavior.

### DIFF
--- a/magenta/models/melody_rnn/BUILD
+++ b/magenta/models/melody_rnn/BUILD
@@ -78,6 +78,7 @@ py_library(
         ":melody_rnn_config",
         ":melody_rnn_graph",
         "//magenta",
+        # numpy dep
         # tensorflow dep
     ],
 )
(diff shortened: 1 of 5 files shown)


jesseengel:
Looks good. It actually improves readability, nice.

iansimon:
I updated this to use the GeneratorOptions to pass beam search parameters.

However, there's an issue with batch size.  If at all possible, I'd really like to require that beam size and batch size are the same; the code will be a lot messier otherwise.  Unfortunately batch size is currently fixed by the graph itself, and it looks nontrivial to remove this.  (It's also fixed at 1 in our existing bundles.)

PR #276, opened 2016-10-18
fix bug where temperature parameter was being ignored

iansimon:


magenta/models/melody_rnn/melody_rnn_sequence_generator.py
```diff
@@ -34,12 +34,16 @@ class MelodyRnnSequenceGeneratorException(Exception):
 class MelodyRnnSequenceGenerator(magenta.music.BaseSequenceGenerator):
   """Shared Melody RNN generation code as a SequenceGenerator interface."""
 
-  def __init__(self, config, steps_per_quarter=4, checkpoint=None, bundle=None):
+  def __init__(self, config, temperature, steps_per_quarter=4, checkpoint=None,
```
adarob: Adding another argument here is going to screw up the interfaces. Can you just add it to the config in the melody_rnn_generate binary instead?

iansimon: What if I just give it a default value of 1.0?

magenta/models/melody_rnn/melody_rnn_generate.py
```diff
@@ -217,6 +217,8 @@ def run_with_flags(generator):
 
 def main(unused_argv):
   """Saves bundle or runs generator based on flags."""
+  config = melody_rnn_config.config_from_flags()
+  config.hparams.temperature = FLAGS.temperature
```
adarob: I think we should set the default for the flag to None and only override if it is given so as to avoid unintentionally overriding the default of the config.

iansimon: I see, makes sense.

PR #277, opened 2016-10-18
Modify sequence generator to take an input section and not require starting from time 0.

adarob:


magenta/music/sequences_lib.py
```diff
@@ -45,6 +44,34 @@ class NegativeTimeException(Exception):
   pass
 
 
+def extract_subsequence(sequence, start_time, end_time):
```
iansimon: Is sequences_lib.py the right file for this function?  I guess it is...

adarob: I don't know if it's the *right* one, but I think it's the *best* and I'd rather not create a new file at this point.

PR #280, opened 2016-10-19
Add optional trigger to magenta midi for starting call phrase.

adarob:


jesseengel:
Code looks good to me

PR #287, opened 2016-10-24
Fix a few bugs in lookback and key-melody encoder-decoders.

iansimon:
Fixes the following two bugs:

1) First lookback one-hot vector is written on top of current note one-hot vector.  What's worse, the test data appears to also reflect the error.

2) In both lookback and key-melody, some offsets into the input vector are hard-coded instead of being dependent on the number of lookback distances & binary counter bits.


magenta/music/melody_encoder_decoder.py
```diff
@@ -361,19 +365,23 @@ def events_to_input(self, events, position):
...
+      input_[(i + 1) * self.num_melody_events + index] = 1.0
+
+    num_lookbacks = len(self._lookback_distances)
 
     # Binary time counter giving the metric location of the *next* note.
     n = position + 1
     for i in range(self._binary_counter_bits):
-      input_[3 * self.num_melody_events + i] = 1.0 if (n / 2 ** i) % 2 else -1.0
+      input_[(num_lookbacks + 1) * self.num_melody_events +
+             i] = 1.0 if (n / 2 ** i) % 2 else -1.0
```
adarob: put the i on the previous line and use parenthesis after the equals sign

adarob: Should we just be using 0 instead of -1.0? @elliotwaite?

iansimon: Fixed.

iansimon: At the very least using -1.0 seems inconsistent with all of the other binary features.

PR #288, opened 2016-10-24
Switch to using steps vs quarters for finer granularity in call and response.

adarob:


magenta/interfaces/midi/midi_interaction.py
```diff
@@ -127,38 +128,41 @@ class CallAndResponseMidiInteraction(MidiInteraction):
...
 
   def run(self):
     """The main loop for a real-time call and response interaction."""
 
-    # We measure time in units of quarter notes.
-    quarter_duration = 60.0 / self._qpm
-    # Start time in quarter notes from the epoch.
-    start_quarters = (time.time() + 1.0) // quarter_duration
+    # We measure time in units of steps.
+    step_duration = 60.0 / (self._qpm * self._steps_per_quarter)
```
cghawthorne: since duration is ambiguous, can you add a unit to this variable name?

adarob: Done.

PR #298, opened 2016-10-26
Add temperature control to magenta_midi.

adarob:


magenta/interfaces/midi/magenta_midi.py
```diff
@@ -45,22 +45,27 @@
     'phrase_bars',
     None,
     'The number of bars of duration to use for the call and response phrases. '
-    'If none, `end_call_control_number` must be specified.')
```
jesseengel: I think actually `control_number` is much clearer than `control` because it could also refer to value. For me hacking things, I found `number` to make things clearer.

adarob: Done.

Issue #282, opened 2016-10-20
Split model and generator


cghawthorne:
Move generate_melody into the model class. Instantiate models using a config object.

PR #306, opened 2016-10-27
Split model and sequence generation code

cghawthorne:
Moves generate_melody into MelodyRnnModel so that MelodyRnnSequenceGenerator is concerned only with acting as a bridge between the sequence generator API and the model. If you only want to generate a melody and don't care about NoteSequence interaction, you can just instantiate the model directly.

Also creates a BaseModel class that defines some standard checkpoint-handling methods and moves that logic out of BaseSequenceGenerator. The goal here is to have the model class be the place where the graph is managed. The sequence generator doesn't need to know anything about the graph.

Finally, updates the error checking logic around empty sequences. It is now the caller's responsibility to supply some kind of priming sequence. If the sequence is silence, the model will just be primed with silence instead of a randomly-generated note (this is the case if you don't play any notes during a midi call/response cycle). If the sequence is completely empty, an exception will be thrown. To prevent this exception when running the command line generation script, added a hard-coded default sequence if no primer is given.

### DIFF
--- a/magenta/models/melody_rnn/BUILD
+++ b/magenta/models/melody_rnn/BUILD
@@ -65,7 +65,9 @@ py_library(
     srcs = ["melody_rnn_model.py"],
     srcs_version = "PY2AND3",
     deps = [
+        ":melody_rnn_graph",
         "//magenta",
+        # tensorflow dep
     ],
 )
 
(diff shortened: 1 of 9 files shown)


cghawthorne:
Fixes #282

PR #314, opened 2016-10-29
Initial import of Kyle's polyphonic RNN model

cghawthorne:
Still to do (for future PRs):
- Add ability to create dataset (depends on the new musicxml importer)
- Wrap the model in a SequenceGenerator interface
- Documentation


cghawthorne:
@kastnerkyle feel free to review if you'd like!

magenta/models/polyphonic_rnn/polyphonic_rnn_lib.py
```diff
@@ -0,0 +1,1377 @@
...
+
+
+##
+# begin metautils
+##
+
+
+def shape(x):
+  # Get shape of Variable through hacky hacky string parsing
+  shape_tup = repr(tf.Print(x, [tf.shape(x)])).split('shape=')[1]
```
kastnerkyle: There must be a better way to get numeric shapes out of TF, but I couldn't find it. We talked about a way, but when I tried it I couldn't get the information I needed. Maybe there is a better way from the public API that I missed?

magenta/models/polyphonic_rnn/polyphonic_rnn_lib.py
```diff
@@ -0,0 +1,1377 @@
...
+  return shape_tup
+
+
+def ndim(x):
+  return len(shape(x))
+
+
+# TODO: How can I correct this for shared / tied weights?
+def print_network(params):
+  n_params = sum([np.prod(shape(p)) for p in params])
```
kastnerkyle: This probably isn't used in the end, I used it for debug mostly. Also it is currently not right (I don't think) given all the shared parameters in the final model

magenta/models/polyphonic_rnn/polyphonic_rnn_lib.py
```diff
@@ -0,0 +1,1377 @@
...
+  tf.logging.info('Total number of parameters: %fM' % (n_params / float(1E6)))
+
+
+def dot(a, b):
+  # Generalized dot for nd sequences, assumes last axis is projection
+  # b must be rank 2
+  # rediculously hacky string parsing... wowie
+  a_tup = shape(a)
```
kastnerkyle: I think @craffel has a nicer/more general "dot", but this one works for this particular model

magenta/models/polyphonic_rnn/polyphonic_rnn_lib.py
```diff
@@ -0,0 +1,1377 @@
...
+
+def duration_and_pitch_to_midi(filename, durations, pitches, prime_until=0):
+  # durations and pitches should both be 2D
+  # [time_steps, n_notes]
+
+  sequence = music_pb2.NoteSequence()
+
+  # Hardcode for now, eventually randomize?
+  # or predict...
+  sequence.ticks_per_quarter = mm.STANDARD_PPQ
```
kastnerkyle: Some of the generation stuff is hardcoded, could be pulled out as parameters to the call. I don't know how much functionality you want to expose to the user though.

magenta/models/polyphonic_rnn/polyphonic_rnn_lib.py
```diff
@@ -0,0 +1,1377 @@
...
+  # this should be moved out for sure - this is really duration class, pitch!
+  time_classes = [
+      0.125, 0.25, 0.5, 0.75, 1.0, 1.5, 2.0, 3.0, 4.0, 6.0, 8.0, 12.0]
+
+  dt = copy.deepcopy(durations)
+  for n in range(len(time_classes)):
+    dt[dt == n] = time_classes[n]
+
+  # why / 8?
+  delta_times = [dt[..., i] / 8 for i in range(sn)]
```
kastnerkyle: I have no jsutification for this /8, but the generations just didn't sound right at /1 /2 /4, so something might be up with durations / ticks per quarter settings.

magenta/models/polyphonic_rnn/polyphonic_rnn_lib.py
```diff
@@ -0,0 +1,1377 @@
...
+  end_times = [delta_times[i].cumsum(axis=0) for i in range(sn)]
+  start_times = [end_times[i] - delta_times[i] for i in range(sn)]
+  voices = [pitches[..., i] for i in range(sn)]
+
+  midi_notes = []
+  default_instrument = 0
+  default_program = 0
+  priming_instrument = 79
```
kastnerkyle: could expose these if people don't want to hear priming as bassoon

magenta/models/polyphonic_rnn/polyphonic_rnn_lib.py
```diff
@@ -0,0 +1,1377 @@
...
+
+      self._min_time_data = np.min(all_ds)
+      self._max_time_data = np.max(all_ds)
+      self.time_classes = list(np.unique(all_ds.ravel()))
+
+      truncate = len(all_ds) - len(all_ds) % minibatch_size
+      all_ds = all_ds[:truncate]
+      all_ps = all_ps[:truncate]
+
+      # transpose necessary to preserve data structure!
```
kastnerkyle: This logic is supposed to cut the audio into long contiguous subsequences based on the minibatch size - I didn't comment it very well

magenta/models/polyphonic_rnn/polyphonic_rnn_graph.py
```diff
@@ -0,0 +1,179 @@
...
+    sequence_length = 30
+
+    self.train_itr = polyphonic_rnn_lib.TFRecordDurationAndPitchIterator(
+        examples, self.batch_size, stop_index=.9,
+        sequence_length=sequence_length)
+
+    duration_mb, note_mb = next(self.train_itr)
+    self.train_itr.reset()
+
+    self.valid_itr = polyphonic_rnn_lib.TFRecordDurationAndPitchIterator(
```
kastnerkyle: Do the other RNN models hold out part of user data as a validation set? Or should we just train on the whole thing? In the extreme could even early stop, then retrain with all data to the same number of epochs. I don't know what the "normal" thing is for Magenta models

kastnerkyle:
Cool! I tried to leave comments on the parts that seem most relevant - you did an awesome job wrapping this. The main parts seem ok, and most (all) of my comments could be addressed in future PRs. Most are related to user interface in sampling, if you even end up using my functions for sampling -> MIDI. There are probably better ones in Magenta

kastnerkyle:
One other thing to think about in future PR is incorporating parts of [Modelling Symbolic Music: Beyond the Piano Roll, C. Walder](https://arxiv.org/abs/1606.01368). See the [samples](http://users.cecs.anu.edu.au/~christian.walder/). As far as I can tell, he uses a groundtruth rhythm which includes time signature, and every note duration and resamples the pitch values. This could potentially be incorporated simply by modifying the mask and sampling code for this polyphonic model, though a better version would just avoid parameterizing the softmax for duration at all. In general, the sounds are good and it seems like this kind of model (or the reverse - predict durations given pitches) could be a fun addition for Magenta

magenta/models/polyphonic_rnn/polyphonic_rnn_lib.py
craffel: I do:
```Python
def int_shape(x):
  """Returns the shape of a tensor as a tuple of integers or None entries.

  Adapted from keras.backend.tensorflow_backend.

  Args:
    x: A TensorFlow tensor object.

  Returns:
    A tuple of integers or None (for unknown dimensions) corresponding to the
    shape of `x`.
  """
  shape = x.get_shape()
  return tuple([i.__int__() for i in shape])


def ndim(x):
  """Returns the number of dimensions of the TensorFlow variable `x`.

  Args:
    x: A TensorFlow tensor object.

  Returns:
    An integer denoting the number of dimensions in `x`.
  """
  return x.get_shape().ndims


def dot(x, y):
  """Computes the dot product of two N-dimensional tensors.

  Uses the same conventions as np.dot, allowing for variable-length dimensions.
  Note that the last dimension of x and the second-to-last dimension of y must
  be equal and known ahead of time.

  Args:
    x: TensorFlow tensor object.
    y: TensorFlow tensor object.

  Returns:
    A TensorFlow tensor which is the dot product of `x` and `y`.
  """
  # Flatten all dimensions of x up to the last
  xt = tf.reshape(x, [-1, int_shape(x)[-1]])
  # Place the second-to-last dimension of y first (this trick from keras)
  y_permute_dim = list(range(ndim(y)))
  y_permute_dim = [y_permute_dim.pop(-2)] + y_permute_dim
  # and flatten all dimensions after the first (which was originally the
  # second-to-last dimension)
  yt = tf.reshape(tf.transpose(y, perm=y_permute_dim), [int_shape(y)[-2], -1])
  # Compute dot product of flattened matrices
  xy = tf.matmul(xt, yt)
  # Reshape to (x_1, x_2, ..., x_{N - 1}, y_1, y_2, ..., y_{M - 3}, y_{M - 1})
  # Using tf.reshape and tf.pack and tf.shape allows for variable dimensions
  xy = tf.reshape(xy, tf.pack([tf.shape(x)[n] for n in range(ndim(x) - 1)] +
                              [tf.shape(y)[n] for n in range(ndim(y) - 2)] +
                              [tf.shape(y)[ndim(y) - 1]]))
```

I think it may also be useful to call `set_shape` after the `reshape` because `reshape` makes all of the dimensions unknown even when they are known.

cghawthorne: @craffel is this available as a library that we should be using? Or should I just copy/paste this in to replace @kastnerkyle's code?

craffel: Not available anywhere, sorry.  I think at some point something like this should be added to TensorFlow as `tf.dot`, or something, but that's just my 2c.

magenta/models/polyphonic_rnn/polyphonic_rnn_lib.py
cghawthorne: Yup. x.get_shape() is what you want. I'll fix it.

magenta/models/polyphonic_rnn/polyphonic_rnn_lib.py
cghawthorne: Removed.

magenta/models/polyphonic_rnn/polyphonic_rnn_lib.py
cghawthorne: @craffel have you tried adding this as a pull request to tensorflow?

magenta/models/polyphonic_rnn/polyphonic_rnn_lib.py
cghawthorne: Yeah, I suspect I'll have to make a lot more things user-controlled when I start adding the sequence generator interface.

magenta/models/polyphonic_rnn/polyphonic_rnn_lib.py
```diff
@@ -0,0 +1,1377 @@
...
+  ks.key = 0
+  ks.mode = ks.MAJOR
+
+  tempos = sequence.tempos.add()
+  tempos.qpm = 120
+  # ti.simultaneous_notes
+  sn = 4
+  # ti.time_classes
+  # this should be moved out for sure - this is really duration class, pitch!
+  time_classes = [
```
cghawthorne: Can you provide some sample code of how to get it from the iterator?

magenta/models/polyphonic_rnn/polyphonic_rnn_lib.py
cghawthorne: Just to make sure I'm understanding this correctly, does /8 mean that the RNN sees the sequence of notes in 8th note steps?

magenta/models/polyphonic_rnn/polyphonic_rnn_lib.py
cghawthorne: Yeah, I'll leave that up to the sequence generator interface.

magenta/models/polyphonic_rnn/polyphonic_rnn_lib.py
cghawthorne: Added your comment.

magenta/models/polyphonic_rnn/polyphonic_rnn_lib.py
```diff
@@ -0,0 +1,1377 @@
...
+    output_tensors.append(masks[i] * input_tensor)
+  return output_tensors
+
+
+def linear(list_of_inputs, input_dims, output_dim, random_state, name=None,
+           init=None, scale='default', weight_norm=None, biases=True):
+  """Can pass weights and biases directly if needed through init."""
+  if weight_norm is None:
+    # Let other classes delegate to default of linear
+    weight_norm = True
```
cghawthorne: I'll leave it for now.

magenta/models/polyphonic_rnn/polyphonic_rnn_lib.py
cghawthorne: Chatted with @craffel offline. I've updated the shape and ndims methods, but I'll leave the dot method as is until there's an official implementation.

magenta/models/polyphonic_rnn/polyphonic_rnn_lib.py
```diff
@@ -0,0 +1,1371 @@
...
+            new_ds_list.append(ds)
+
+            new_ps_list.append(new_down)
+            new_ds_list.append(ds)
+        all_ds = np.concatenate(new_ds_list)
+        all_ps = np.concatenate(new_ps_list)
+
+      self._min_time_data = np.min(all_ds)
+      self._max_time_data = np.max(all_ds)
+      self.time_classes = list(np.unique(all_ds.ravel()))
```
cghawthorne: What about this as an approach: hard code a set of acceptable durations (possibly the one you already have for bach) and just skip any input that has a duration not in that set. This should work with any kind of input and won't cause problems if training and eval sets have different sets of durations.

kastnerkyle:
That is probably the best approach. The list of valid durations can always
increase later if people are running into a surprising number of unknown
durations.

kastnerkyle:
So it turns out email reply doesn't reply to the commit comment directly...

¯\_(ツ)_/¯

cghawthorne:
@kastnerkyle 555a2b7 implements the time class change you suggested. Can you take a look and see if that matches with what you were thinking?

Are there any other changes you think should happen before this initial checkin?

kastnerkyle:
Looks good to me - this seems like plenty for a first pass. This is great!

PR #319, opened 2016-11-01
Avoid pulling in magenta deps when building pip package

### LINKED ISSUE #317: Building the pip package shouldn't require installing Magenta dependencies
Currently, building the pip package requires installing Magenta dependencies like pretty_midi because setup.py does an `import magenta` to get the version number. We should move the version number into a separate file (version.py?) that is then referenced by both magenta/__init__.py and setup.py

cghawthorne:
Fixes #317

Also regenerate the main magenta dependencies.

### DIFF
--- a/magenta/pipelines/BUILD
+++ b/magenta/pipelines/BUILD
@@ -135,8 +135,6 @@ py_library(
     srcs = ["pipelines_common.py"],
     deps = [
         ":pipeline",
-        "//magenta/music:chord_symbols_lib",
-        "//magenta/music:chords_lib",
         "//magenta/music:melodies_lib",
         "//magenta/music:sequences_lib",
         "//magenta/protobuf:music_py_pb2",
--- a/magenta/pipelines/BUILD
+++ b/magenta/pipelines/BUILD
@@ -135,7 +135,6 @@ py_library(
     srcs = ["pipelines_common.py"],
     deps = [
         ":pipeline",
-        "//magenta/music:melodies_lib",
         "//magenta/music:sequences_lib",
         "//magenta/protobuf:music_py_pb2",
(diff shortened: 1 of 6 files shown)


cghawthorne:
This is a little hacky, so let me know if you have thoughts about a better way to do it.

magenta/pipelines/BUILD
```diff
@@ -135,8 +135,6 @@ py_library(
     srcs = ["pipelines_common.py"],
     deps = [
         ":pipeline",
-        "//magenta/music:chord_symbols_lib",
-        "//magenta/music:chords_lib",
         "//magenta/music:melodies_lib",
```
iansimon: You can also get rid of the melodies_lib dependency; not sure how I missed all of these.

magenta/tools/pip/setup.py
```diff
@@ -16,7 +16,10 @@
 from setuptools import find_packages
 from setuptools import setup
 
-import magenta
+# Bit of a hack to parse the version string stored in version.py without
+# executing __init__.py, which will end up requiring a bunch of dependencies to
+# execute (e.g., tensorflow, pretty_midi, etc.).
+eval(compile(open('magenta/version.py').read(), 'magenta/version.py', 'exec'))
```
iansimon: This is similar to #3 here: https://packaging.python.org/single_source_version/

magenta/pipelines/BUILD
cghawthorne: Done

magenta/tools/pip/setup.py
cghawthorne: Oh neat, I didn't know about execfile. Switched to using that.

PR #326, opened 2016-11-01
Updating models readme to include image stylization 

fredbertsch:


magenta/models/README.md
```diff
@@ -2,6 +2,12 @@
 
 This directory contains Magenta models.
 
+## Image stylization
+
+This is the Multistyle Pastiche Generator model described in
```
cghawthorne: might as well add a link to the model here

fredbertsch:
Done. I link to the directory rather than a specific script since there's a few options.

PR #337, opened 2016-11-01
require tensorflow 0.11.0rc1 instead of 0.11.0rc2

iansimon:


cghawthorne:
I think we can leave it at 11.0rc2. The only problem is with the docker image, and I was able to rely on the 11.0rc1 docker image, and then let our pip package end up installing 11.0rc2. So I think it all works out.

Hopefully the Bazel problem gets fixed soon.

iansimon:
Ah ok, will close.

Issue #341, opened 2016-11-02
Missing flute_scale_with_png.mxl file


cghawthorne:
I'm seeing a Bazel test failure related to flute_scale_with_png.mxl not being found. @jsawruk, did you maybe forget to check that one in?

jsawruk:
Oops, forgot to check it in. It's now on the musicxml branch, commit 0eb36d5.

cghawthorne:
Cool. Can you do a new pull request for that?

jsawruk:
Pull request #342.

cghawthorne:
Looks like the pull request was closed before it could be merged. Were you planning on making a different pull request?

jsawruk:
I didn't like how the rebase went, so I deleted the branch, rebranched off of master, and created a new branch. I guess I need to create a new pull request now.

PR #342, opened 2016-11-02
Fix Issue #341: Missing flute_scale_with_png.mxl file

jsawruk:
Forgot to add this test file in pull request #214. This file has now been added. It is a compressed MusicXML file with two rootfiles: the uncompressed MusicXML file and a PNG file of the notation.


cghawthorne:
Can you rebase your change onto master so it has only the last commit?

Or alternatively, resync with magenta/master and create a new branch that has only the commit to add the file.

PR #347, opened 2016-11-03
MusicXML parser: Stop adding an extra key signature at time -1

cghawthorne:
Also adds a unit test to verify the contents of the flute scale.


cghawthorne:
FYI @jsawruk

jsawruk:
Thanks for fixing this, as well as demonstrating a different way to do unit testing.

PR #348, opened 2016-11-03
git pull to fetch latest changes.

johndpope:


johndpope:
this has a hard git commit in dockerfile that I think is causing problems with updated code.

 https://github.com/cghawthorne/magenta/blob/246e8ea3483fc90dcd3c5c4347ca0bcb64156e1a/magenta/tools/docker/Dockerfile

PR #351, opened 2016-11-03
Update docker example command

### LINKED ISSUE #350: Exactly one of `--config` or `--melody_encoder_decoder` must be supplied.
dockerfile 

running this inside docker
```
melody_rnn_generate \
  --bundle_file=/magenta-models/lookback_rnn.mag \
  --output_dir=/magenta-data/lookback_rnn/generated \
  --num_outputs=10 \
  --num_steps=128 \
  --primer_melody="[60]"

```
 -> blows up 

screen shot 2016-11-03 at 16 02 21

tried passing

`
 melody_rnn_generate   --bundle_file=/magenta-models/lookback_rnn.mag   --output_dir=/magenta-data/lookback_rnn/generated   --num_outputs=10   --num_steps=128   --primer_melody="[60]" --melody_encoder_decoder='lookback'
`

```
melody_encoder_decoders = {
    'onehot': one_hot_melody_encoder_decoder,
    'lookback': lookback_melody_encoder_decoder,
    'key': magenta.music.KeyMelodyEncoderDecoder
}
```

screen shot 2016-11-03 at 16 06 49
but fails

AttributeError: 'NoneType' object has no attribute 'id'

I'm using the dockerfile from the readme.

cghawthorne:
Needs to include the --config parameter for melody_rnn_generate.

Fixes #350

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,6 +99,7 @@ The Docker image also includes several pre-trained models in
 
 ```
 melody_rnn_generate \
+  --config=lookback_rnn \
   --bundle_file=/magenta-models/lookback_rnn.mag \
   --output_dir=/magenta-data/lookback_rnn/generated \
   --num_outputs=10 \


cghawthorne:
Yeah, I think we should figure out some way of making that automatic so users don't have to specify the config when they're loading a bundle.

johndpope:
I'm eager to test this out in Docker

It seems Dockerfile is hard coded to a specific commit on magenta (MAGENTA_MASTER_REF) and tensorflow. Is it not possible to have it fetch latest code from master?
I tried rebuilding container from scratch - drafted here (including other midi dependencies)
https://gist.github.com/johndpope/2d3a90c8d70b22732ad75b6ebec60187

but eventually fails with 
/magenta/magenta/music/BUILD:44:1: no such package '@music21//': File "LICENSE" linked from "LICENSE" does not exist and referenced by '//magenta/music:chord_symbols_lib'.
ERROR: Analysis of target '//magenta/music:chord_symbols_lib_test' failed; build aborted.

ok from some digging it seems like this is from a requirement of bazel 0.3.1

cghawthorne:
@johndpope, we build new Docker images with each major release. The Docker image contains a full development environment, so you could certainly update to the latest git commit within the image, but unless you moved the git repo to a persistent volume, your updates would disappear when your session ended.

If you'd like to experiment with the latest changes from master, I'd recommend setting up a local development environment: https://github.com/tensorflow/magenta#development-environment

Also, you're right that we require Bazel 0.3.1 for now. There's a bug in 0.3.2 that breaks out build that we're hoping will be resolved soon. You can follow updates here: https://github.com/bazelbuild/bazel/issues/1997

johndpope:
Thanks Craig.

Issue #354, opened 2016-11-04
MusicXML Parser: Transposition is broken


jsawruk:
When trying to read in a transposing MusicXML file I got this error:

```
musicxml_parser.py", line 413, in _parse_attributes
    self.key_signature.key += transpose
AttributeError: 'NoneType' object has no attribute 'key'
```

It appears to be a regression bug from pull request #347.

cghawthorne:
Thanks for catching this. Could you investigate and add a unit test to verify the fix?

cghawthorne:
Thanks, @jsawruk!

PR #355, opened 2016-11-04
MusicXML: Fix transposition bug when no key signature found

jsawruk:
Fix issue #354.

Only changed the transposition of the key signature if a key signature was present. If the music has no key, but the part changes transposition (i.e. a change of instruments within a part), the key signature remains none.

### DIFF
--- a/magenta/music/musicxml_parser.py
+++ b/magenta/music/musicxml_parser.py
@@ -410,7 +410,8 @@ class Measure(object):
       elif child.tag == 'transpose':
         transpose = int(child.find('chromatic').text)
         self.state.transpose = transpose
-        self.key_signature.key += transpose
+        if self.key_signature != None:
+          self.key_signature.key += transpose
       else:
         # Ignore other tag types because they are not relevant to Magenta.
         pass
--- a/magenta/music/musicxml_parser.py
+++ b/magenta/music/musicxml_parser.py
@@ -410,7 +410,7 @@ class Measure(object):
       elif child.tag == 'transpose':
         transpose = int(child.find('chromatic').text)
         self.state.transpose = transpose
-        if self.key_signature != None:
+        if self.key_signature is not None:
           self.key_signature.key += transpose
       else:
         # Ignore other tag types because they are not relevant to Magenta.
--- a/magenta/music/musicxml_parser_test.py
+++ b/magenta/music/musicxml_parser_test.py
@@ -342,6 +342,11 @@ class MusicXMLParserTest(tf.test.TestCase):
         }
         key_signatures: {
         }
+        source_info: {
+          source_type: SCORE_BASED
+          encoding_type: MUSIC_XML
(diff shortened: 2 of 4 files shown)


magenta/music/musicxml_parser_test.py
```diff
@@ -300,5 +313,40 @@ def testFluteScale(self):
...
+  def testatonaltransposition(self):
+    """Test that transposition works when changing instrument transposition
+    within a single part in a score that has no key signature / is atonal.
+    """
+    ns = musicxml_reader.musicxml_file_to_sequence_proto(
+        self.atonal_transposition_filename)
+    expected_ns = common_testing_lib.parse_test_proto(
+        music_pb2.NoteSequence,
+        """
+        ticks_per_quarter: 220
```
cghawthorne: You'll need to sync with master and add the following lines here due to #353:
```
 source_info {
   source_type: SCORE_BASED
   encoding_type: MUSIC_XML
   parser: MAGENTA_MUSIC_XML
 }
```

magenta/music/musicxml_parser.py
```diff
@@ -410,7 +410,8 @@ def _parse_attributes(self, xml_attributes):
       elif child.tag == 'transpose':
         transpose = int(child.find('chromatic').text)
         self.state.transpose = transpose
-        self.key_signature.key += transpose
+        if self.key_signature != None:
```
cghawthorne: We prefer `if self.key_signature is not None`

https://google.github.io/styleguide/pyguide.html?showone=True/False_evaluations#True/False_evaluations

magenta/music/musicxml_parser_test.py
```diff
@@ -300,5 +313,40 @@ def testFluteScale(self):
     self.maxDiff = None
     self.assertProtoEquals(expected_ns, ns)
 
+  def testatonaltransposition(self):
+    """Test that transposition works when changing instrument transposition
```
cghawthorne: function docstrings should have a summary line that ends with ".", "?", or "!", followed by a blank line, and then followed by any additional information.

https://google.github.io/styleguide/pyguide.html?showone=Comments#Comments

magenta/music/musicxml_parser_test.py
```diff
@@ -300,5 +313,40 @@ def testFluteScale(self):
     self.maxDiff = None
     self.assertProtoEquals(expected_ns, ns)
 
+  def testatonaltransposition(self):
```
cghawthorne: function names should be lowercase with words separated by underscores: `def test_atonal_transposition`. We've been a little inconsistent with that ourselves, but we should make sure new methods follow the correct pattern.

https://google.github.io/styleguide/pyguide.html?showone=Comments#Naming

magenta/music/musicxml_parser_test.py
cghawthorne: This is still missing the source_info fields.

PR #356, opened 2016-11-04
Extract part and voice information from MusicXML files

cghawthorne:
- Add a test to verify new St. Anne test file with 2 parts and 4 voices.
- Switched parser to linking parts and score-part elements based on id, rather than the order they appeared in the document.
- If no midi channel is indicated in the ScorePart, just use the default midi channel, rather than using a new channel for every part. There only 16 midi channels, but an unlimited number of parts.
- Made some fields/methods in musicxml_parser private that didn't need to be public.

### DIFF
--- a/magenta/music/musicxml_parser.py
+++ b/magenta/music/musicxml_parser.py
@@ -95,7 +95,7 @@ class MusicXMLDocument(object):
   """Internal representation of a MusicXML Document.
 
   Represents the top level object which holds the MusicXML document
-  Responsible for loading the .xml or .mxl file using the get_score method
+  Responsible for loading the .xml or .mxl file using the _get_score method
   If the file is .mxl, this class uncompresses it
 
   After the file is loaded, this class then parses the document into memory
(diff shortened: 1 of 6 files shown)


cghawthorne:
@jsawruk would appreciate your MusicXML expertise on this review too.

magenta/music/musicxml_parser.py
```diff
@@ -178,24 +178,20 @@ def get_score(filename):
...
 
     # Parse parts
-    for score_part_index, child in enumerate(self.score.findall('part')):
-      # If a score part is missing, add a default score part
-      if score_part_index >= len(self.score_parts):
-        score_part = ScorePart(self.state)
-        self.score_parts.append(score_part)
-
-      part = Part(child, self.score_parts[score_part_index], self.state)
+    for score_part_index, child in enumerate(self._score.findall('part')):
```
iansimon: What is score_part_index for?

magenta/music/musicxml_parser.py
```diff
@@ -293,37 +291,34 @@ class ScorePart(object):
...
     if xml_score_part is not None:
-      self._parse()
+      self._parse(xml_score_part)
 
-  def _parse(self):
+  def _parse(self, xml_score_part):
     """Parse the <score-part> element to an in-memory representation."""
-    if self.xml_score_part.find('part-name') is not None:
-      self.part_name = self.xml_score_part.find('part-name').text
+    self.id = xml_score_part.attrib['id']
```
iansimon: Is there always an id?

jsawruk: Yes, it is required by the MusicXML specification: http://usermanuals.musicxml.com/MusicXML/MusicXML.htm#EL-MusicXML-part.htm

magenta/music/musicxml_parser.py
jsawruk: There is supposed to be a 1:1 between ScorePart (MIDI info like track, patch, etc.) and Part, but I have found some (particularly older) MusicXML files where this not true. If a Part is found but there isn't a ScorePart, the parser creates a default ScorePart with default MIDI patch. The score_part_index then is used to associate the correct ScorePart with the Part.

I know this is confusing, and I would have preferred it if the MIDI info was contained directly in the Part info and that there wasn't a ScorePart element. I'm open to suggestions on better ways to handle this. For example, we may want to raise an Exception if there is a missing ScorePart, rather than try to correct bad data. This might simplify the code, but I do have a lot of older MusicXML that seem to have this encoding error.

PR #359, opened 2016-11-06
add flag exluding primer sequence

keypulsations:
Adds a flag, `exclude_primer_midi`, to `melody_rnn_generate.py` that, if set to `True`, results in removing notes in generated_sequence that start before the calculated start_time in the generate_section request.

### DIFF
--- a/magenta/models/melody_rnn/melody_rnn_generate.py
+++ b/magenta/models/melody_rnn/melody_rnn_generate.py
@@ -97,6 +97,10 @@ tf.app.flags.DEFINE_string(
     'log', 'INFO',
     'The threshold for what messages will be logged DEBUG, INFO, WARN, ERROR, '
     'or FATAL.')
+tf.app.flags.DEFINE_boolean(
+    'exclude_primer_midi', True,
+    'If true, instead of including the primer MIDI in the generated sequence, '
+    'will only output the newly generated MIDI.')
 
 
 def get_checkpoint():
@@ -194,9 +198,11 @@ def run_with_flags(generator):
     # Set the start time to begin on the next step after the last note ends.
     last_end_time = (max(n.end_time for n in primer_sequence.notes)
                      if primer_sequence.notes else 0)
+    start_time = last_end_time + _steps_to_seconds(1, qpm)
+
     generate_section = generator_options.generate_sections.add(
-        start_time=last_end_time + _steps_to_seconds(1, qpm),
-        end_time=total_seconds)
+      start_time=start_time,
+      end_time=total_seconds)
 
     if generate_section.start_time >= generate_section.end_time:
       tf.logging.fatal(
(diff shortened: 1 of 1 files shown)


keypulsations:
I signed it!

magenta/models/melody_rnn/melody_rnn_generate.py
```diff
@@ -194,9 +198,11 @@ def run_with_flags(generator):
     # Set the start time to begin on the next step after the last note ends.
     last_end_time = (max(n.end_time for n in primer_sequence.notes)
                      if primer_sequence.notes else 0)
+    start_time = last_end_time + _steps_to_seconds(1, qpm)
+
     generate_section = generator_options.generate_sections.add(
-        start_time=last_end_time + _steps_to_seconds(1, qpm),
-        end_time=total_seconds)
+      start_time=start_time,
```
cghawthorne: we use 4 spaces for line continuations, so you'll need to add another 2 spaces before these 2 lines.

magenta/models/melody_rnn/melody_rnn_generate.py
```diff
@@ -223,9 +229,23 @@ def run_with_flags(generator):
   # files.
   date_and_time = time.strftime('%Y-%m-%d_%H%M%S')
   digits = len(str(FLAGS.num_outputs))
+
   for i in range(FLAGS.num_outputs):
     generated_sequence = generator.generate(input_sequence, generator_options)
 
+    if FLAGS.exclude_primer_midi:
+      # Remove the first half of the sequence. Since we are continually
```
cghawthorne: Depending on the `num_steps` flag and the length of the primer, removing the first half may not be correct. I would recommend removing notes from `generated_sequence` only if their `note.start_time` is earlier than `start_time`.

keypulsations:
@cghawthorne thanks for your feedback. Please let me know how this looks!

magenta/models/melody_rnn/melody_rnn_generate.py
```diff
@@ -223,9 +229,21 @@ def run_with_flags(generator):
...
   for i in range(FLAGS.num_outputs):
     generated_sequence = generator.generate(input_sequence, generator_options)
 
+    if FLAGS.exclude_primer_midi:
+      # Remove any notes with start_time less than start_time of generated
+      # sequence. Otherwise, subtract length of removed input_sequence from
+      # start_time and end_time of remaining notes.
+      for note in generated_sequence.notes[:]:
+        if note.start_time < start_time:
+          generated_sequence.notes.pop(0)
```
keypulsations: @cghawthorne since `note` is already accessible, how about using `remove`:

`generated_sequence.notes.remove(note)`

When using an index, it seems we'll need to enumerate the `generate_sequence.notes` slice, like: 

```
      for index, note in enumerate(generated_sequence.notes[:]):
        if note.start_time < start_time:
          del generated_sequence.notes[index]
```

cghawthorne: Yeah, I think `remove` should be fine too. `del` might be slightly more performant because it won't have to search for the object to remove, but in this context, it shouldn't matter.

keypulsations:
@cghawthorne thanks again for your feedback. I've used `del` to delete any notes in `primer_sequence` if the `exclude_primer_midi` flag is set to `True`. Notes are determined to be in `primer_sequence` if `note.start_time` is `<` `start_time`, which is the starting point of newly generated notes (`start_time = last_end_time + _steps_to_seconds(1, qpm)`).

magenta/models/melody_rnn/melody_rnn_generate.py
```diff
@@ -223,9 +229,21 @@ def run_with_flags(generator):
...
   for i in range(FLAGS.num_outputs):
     generated_sequence = generator.generate(input_sequence, generator_options)
 
+    if FLAGS.exclude_primer_midi:
+      # Delete any notes with start_time less than start_time of generated
+      # sequence. Otherwise, subtract length of removed input_sequence from
+      # start_time and end_time of remaining notes.
+      for index, note in enumerate(generated_sequence.notes[:]):
+        if note.start_time < start_time:
+          del generated_sequence.notes[index]
```
keypulsations: Thanks @cghawthorne. Ok, so maybe `remove` would be the right approach here even if it's more costly. I'll take a look at sequences_lib.extract_subsequence. Appreciate it!

AndrewRahman:
Would _really_ love to see this implemented guys. Can we fix the issue?

PR #363, opened 2016-11-08
Fix time corruption bug

cghawthorne:


cghawthorne:
@kastnerkyle I think this explains why durations needed to be divided by 8.

kastnerkyle:
Yep that looks like it would be bad news - views are deceptive! Thanks @cghawthorne for fixing it

cghawthorne:
The numpy views are a problem, though I think with the way the iterators are currently used, they won't cause issues (though I could be wrong there).

The main problem is that the time class translation searches over data that it has already changed. So for example: it replaces all instances of .5 (seconds) with 1 (class), and then later replaces 1 (seconds) with 4 (class), then 4 (seconds) with 8 (class), then 8 (seconds) with 10 (class). Which means that all instances of .5 seconds get turned into 8 seconds. Exactly how this bug manifests depends on what class values also show up as duration values in a class index larger than the original one.

magenta/models/polyphonic_rnn/polyphonic_rnn_lib.py
```diff
@@ -199,14 +199,17 @@ def duration_and_pitch_to_midi(filename, durations, pitches, prime_until=0):
   sn = 4
 
   # Translate durations from TIME_CLASSES indexes to TIME_CLASSES values.
+  # First, find the data to change, then change it.
+  # If we don't do this as separate steps, we'll end up changing some data
+  # more than once.
   dt = copy.deepcopy(durations)
-  for i, time_class in enumerate(TIME_CLASSES):
-    dt[dt == i] = time_class
+  time_class_indexes = []
```
iansimon: indices?

cghawthorne: indexes is more common in North America: http://grammarist.com/usage/indexes-indices/

PR #365, opened 2016-11-09
Set up RL Tuner in the pip package

cghawthorne:


cghawthorne:
@natashamjaques

cghawthorne:
I removed the `import magenta as mg` line.

The `RL_CODE_PATH` was so it could work without the pip package. I removed that line in this PR because this should be able to use the code directly.

PR #377, opened 2016-11-10
Integrate MusicXML Parser

cghawthorne:
Consolidate MusicXML and MIDI conversion into one script
Add MusicXML conversion APIs to the core Magenta API


cghawthorne:
FYI @jsawruk

yuchima:
The one script seems not working, convert_midi_dir_to_note_sequences it works, but covert_dir_to_note_sequences not working. And convert musicmxl seems not working either.

cghawthorne:
This should be fixed when we cut a new pip package. For now, to use the musicxml converter, you'll probably need to build from the dev environment. We plan on releasing a new pip package soon.

Issue #357, opened 2016-11-04
MusicXML Parser: Add ability to indicate partial measures (e.g., pickup measure)


cghawthorne:
We would eventually like to train models to understand what kinds of things are more common on different beats within measures. For example, it might learn that certain harmonies are more common on "strong" beats (e.g., 1 & 3). In order to do this, we would need to know where notes are within a measure.

For pieces with no partial measures, we have all the information we need. The combination of tempo, time signature, and note duration is sufficient to determine which measure a note is in and what beat within the measure (use tempo to calculate how long a quarter note is, use time signature to determine how many of those durations are in a measure).

However, for pieces with a pickup measure at the beginning (or other kinds of partial measures), this calculation will always be incorrectly offset by the duration of the pickup measure. For example, if a piece is in 4/4 and has a 1-beat pickup measure, we would incorrectly assume that the first 4 beats make up the first measure, when it is actually the first 4 beats _after_ the first pickup beat.

To solve this problem, we should add in new TimeSignature messages to indicate implicit time signature changes within pieces. For example, if a 4/4 piece has a 1-beat pickup measure, several full measures, and a 3-beat final measure, it would have 3 TimeSignature messages: 1/4 for the first measure, 4/4 after that, and a final 3/4 time signature at the end.

cghawthorne:
@jsawruk is this something you would be interested in?

jsawruk:
I think that's doable because we know the cumulative duration within a single measure. For example, I generated a MusicXML file in 4/4 with a 1 beat pickup. In measure 0, it set divisions to 256 and had a note of duration 256. In measure 1, divisions continued as 256, and the whole note had a duration of 1024. (Please see attached example file).

This demonstrates that any partial measure would just be the maximum of the sum of the note durations across all voices with that measure (some voices may only appear partially within a given measure). In the above example, if the durations summed to 512, then a 2/4 TimeSignature would be inserted. 

MusicXML does not enforce that the durations fit within a measure, so we can expand this if a measure overflows, i.e. there are 5 beats in a measure of 4/4.
[pickup_bar.xml.zip](https://github.com/tensorflow/magenta/files/572919/pickup_bar.xml.zip)

jsawruk:
Here's a proposed solution:
- When beginning a new measure, check if there is a previous measure. 
- If there is, find the sum of its durations in voice 1 (since voice 1 is always present)
- Build a time signature as proper fraction in lowest terms (ex: 3/8 instead of 1.5/4)
[...]

PR #380, opened 2016-11-11
MusicXML Parser: Add ability to indicate partial measures 

jsawruk:
Issue #357

### DIFF
--- a/magenta/music/musicxml_parser.py
+++ b/magenta/music/musicxml_parser.py
@@ -95,7 +95,7 @@ class MusicXMLParserState(object):
     # Keep track of current transposition level in +/- semitones.
     self.transpose = 0
 
-    # Keep track of current time signature. Does not support polymeter
+    # Keep track of current time signature. Does not support polymeter.
     self.time_signature = None
 
 
@@ -378,8 +378,6 @@ class Part(object):
     xml_measures = xml_part.findall('measure')
     for child in xml_measures:
       measure = Measure(child, self._state)
-      # Update the time signature if a partial or pickup measure
-      measure.check_time_signature()
       self.measures.append(measure)
 
   def __str__(self):
@@ -405,6 +403,8 @@ class Measure(object):
     # can be inserted at the beginning of the measure
     self.start_time_position = self.state.time_position
     self._parse()
+    # Update the time signature if a partial or pickup measure
+    self._check_time_signature()
 
   def _parse(self):
     """Parse the <measure> element."""
@@ -501,14 +501,16 @@ class Measure(object):
                * self.state.seconds_per_quarter)
     self.state.time_position += seconds
 
-  def check_time_signature(self):
+  def _check_time_signature(self):
     """Correct the time signature for incomplete measures.
 
     If the measure is incomplete or a pickup, insert an appropriate
     time signature into this Measure.
     """
     # Compute the fractional time signature (duration / divisions)
-    fractional_time_signature = Fraction(self.duration, self.state.divisions)
+    # Multiply divisions by 4 because division is always parts per quarter note
+    fractional_time_signature = Fraction(self.duration,
+                                         self.state.divisions * 4)
 
     if self.state.time_signature is None and self.time_signature is None:
       # No global time signature yet and no time signature defined
@@ -516,9 +518,6 @@ class Measure(object):
       # Insert the fractional time signature as the time signature
       # for this measure
 
-      # If no time signature is given, assume an underlying quarter note pulse
-      fractional_time_signature = Fraction(self.duration,
-                                           self.state.divisions * 4)
       self.time_signature = TimeSignature(self.state)
       self.time_signature.numerator = fractional_time_signature.numerator
       self.time_signature.denominator = fractional_time_signature.denominator
@@ -529,11 +528,8 @@ class Measure(object):
       if self.state.time_signature is None:
         self.state.time_signature = self.time_signature
 
-      # Multiply by a time signature of
-      # 1 / (current time signature denominator)
+      # Get the current time signature denominator
       global_time_signature_denominator = self.state.time_signature.denominator
-      fractional_time_signature *= Fraction(1,
-                                            global_time_signature_denominator)
 
       # If the fractional time signature = 1 (e.g. 4/4),
       # make the numerator the same as the global denominator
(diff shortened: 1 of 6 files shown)


magenta/music/musicxml_parser.py
```diff
@@ -465,6 +480,58 @@ def _parse_forward(self, xml_forward):
...
     self.state.time_position += seconds
 
+  def check_time_signature(self):
+    """Correct the time signature for incomplete measures.
+
+    If the measure is incomplete or a pickup, insert an appropriate
+    time signature into this Measure.
+    """
+    # Compute the fractional time signature (duration / divisions)
+    fractional_time_signature = Fraction(self.duration, self.state.divisions)
```
cghawthorne: it looks like this value is used only in the `else` block. might make sense to move it there.

magenta/music/musicxml_parser.py
```diff
@@ -358,6 +361,8 @@ def _parse(self, xml_part, score_parts):
     xml_measures = xml_part.findall('measure')
     for child in xml_measures:
       measure = Measure(child, self._state)
+      # Update the time signature if a partial or pickup measure
+      measure.check_time_signature()
```
cghawthorne: rather than requiring check_time_signature to be called after Measure is constructed, should it be called by the Measure constructor after parse? If so, it should be renamed to `_check_time_signature`.

magenta/music/musicxml_parser.py
```diff
@@ -395,6 +405,11 @@ def _parse(self):
         self.notes.append(note)
         # Keep track of current note as previous note for chord timings
         self.state.previous_note = note
+
+        # Sum up the MusicXML durations in voice 1 of this measure
```
cghawthorne: will this be inaccurate if voice 1 doesn't have any notes? would it be better to determine the duration based on the last note the measure of any voice?

magenta/music/musicxml_parser.py
```diff
@@ -90,6 +90,9 @@ def __init__(self):
     # Keep track of current transposition level in +/- semitones.
     self.transpose = 0
 
+    # Keep track of current time signature. Does not support polymeter
```
cghawthorne: add a period at the end of this comment

magenta/music/musicxml_parser.py
```diff
@@ -465,6 +480,58 @@ def _parse_forward(self, xml_forward):
...
+
+    if self.state.time_signature is None and self.time_signature is None:
+      # No global time signature yet and no time signature defined
+      # in this measure (no time signature or senza misura).
+      # Insert the fractional time signature as the time signature
+      # for this measure
+
+      # If no time signature is given, assume an underlying quarter note pulse
+      fractional_time_signature = Fraction(self.duration,
+                                           self.state.divisions * 4)
```
cghawthorne: why does this calculation multiply self.state.divisions by 4 where the one on line 490 does not?

magenta/music/musicxml_parser.py
```diff
@@ -374,8 +379,13 @@ def __init__(self, xml_measure, state):
     self.tempos = []
     self.time_signature = None
```
cghawthorne: just to verify, is it possible to have multiple time signatures in one measure?

magenta/music/musicxml_parser.py
jsawruk: Yes, that makes more sense.

magenta/music/musicxml_parser.py
jsawruk: That's a very complicated question. Here are some potential answers:
- According to the MusicXML DTDs, it is possible for a time signature to change at any time within a measure. In common practice notation, this is extremely unusual. A time signature only occurs at the beginning of a measure, and a new time signature indicates the start of a new measure, with the exception of a courtesy time signature in a composite time signature (see below).
- It is possible to have a composite time signature such as 2/4 + 3/8 at the beginning of a measure. My code currently does not support this. 
- It is possible to have alternating time signatures, both defined at the beginning of a measure, such as 3/4 6/8. My code currently does not support this. In both this and the previous case, my code will currently only pick up the last time signature.
- It is also possible to have different time signatures within the same measure across different parts (polymeter). Since no major notation program currently supports this, I am not supporting it either.

Do you have any suggestions on how to proceed? All of these cases are edge cases; it is generally safe to assume one time signature per measure in 95%+ of music. I'm not even sure how notation software would encode this because engravers often have to use workarounds to get these edge cases to display, so the output might not be reliable or consistent.

magenta/music/musicxml_parser.py
jsawruk: It is used in both the if and the else blocks. I am simply initializing it here so that the variable has scope in both the if and else.

magenta/music/musicxml_parser.py
jsawruk: I checked with Michael Good, and divisions are always parts per quarter note, so the multiplication by four should exist in both places.

magenta/music/musicxml_parser.py
cghawthorne: Yeah, I think it's fine not to support these edge cases. To make it clear what we do support, I'd suggest adding checks for these circumstances and raising an exception if one occurs.

jsawruk: Would you like an exception to be raised whenever a second time signature in a measure is found? I'll also need to construct a test case for this.

cghawthorne:
Hey @jsawruk, not sure where this ended up with the last round of commits. Is it ready for review again at this point?

jsawruk:
Yes I believe it is ready for another round of review.

magenta/music/musicxml_parser.py
```diff
@@ -486,6 +514,55 @@ def _parse_forward(self, xml_forward):
                * self.state.seconds_per_quarter)
     self.state.time_position += seconds
 
+  def _check_time_signature(self):
```
cghawthorne: could you rename this to _fix_time_signature? I think that makes it a little more clear that this method actually modifies things.

cghawthorne:
I went ahead and corrected a few minor lint fixes. Just have one more comment about a method name change, and then I think we're ready to submit.

cghawthorne:
Thanks!

PR #388, opened 2016-11-14
Fix hparam updates.

adarob:


cghawthorne:
Should add a unit test to verify.

PR #415, opened 2016-11-26
conda now has a package for tensorflow

Mistobaan:
This might be easier to use and copy / paste for quick setup

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,7 +36,7 @@ Jupyter notebooks, and the Magenta scripts are installed in your path!
 Note that you will need to run `source activate magenta` to use Magneta every
 time you open a new terminal window.
 
-#### Manual Install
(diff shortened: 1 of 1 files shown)


cghawthorne:
I'm a little reluctant to switch to the version on conda-forge. It's community-maintained, so it's not always up to date (right now, it's on 0.11.0rc2, but the current version is 0.11.0), and I believe it doesn't have GPU support.

@iansimon what do you think?

iansimon:
I'm somewhat indifferent, but in general I guess I prefer having fewer installation methods to support.

cghawthorne:
Sounds good. For now, I'll close this. I'm hoping with some pip package changes we can automate more of the installation process in the future.

PR #416, opened 2016-11-28
Update rl_tuner for basic_rnn compatibility.

adarob:


adarob:
@natashamjaques please have a look also

PR #427, opened 2016-12-01
Fix model reuse bug

cghawthorne:
Generate a new model instance for every generator constructed.


adarob:
Wrong PR description?

cghawthorne:
Bleh, wrong branch.

PR #451, opened 2016-12-13
Save v1 checkpoints for bundle files

cghawthorne:
This is needed until we update bundle files to support v2 checkpoints (#450)

### DIFF
--- a/magenta/music/model.py
+++ b/magenta/music/model.py
@@ -82,7 +82,7 @@ class BaseModel(object):
       checkpoint_filename: Path to the checkpoint file.
     """
     with self._session.graph.as_default():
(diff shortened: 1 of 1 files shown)


magenta/music/model.py
```diff
@@ -82,7 +82,7 @@ def write_checkpoint_with_metagraph(self, checkpoint_filename):
       checkpoint_filename: Path to the checkpoint file.
     """
     with self._session.graph.as_default():
-      saver = tf.train.Saver(sharded=False)
+      saver = tf.train.Saver(sharded=False, write_version=1)
```
adarob: tf.train.SaverDef.V1

Issue #429, opened 2016-12-01
Magenta on window?


rupeshs:
Happy that tensor flow for windows released but what about magenta will it work  on windows?

Banus:
Currently Tensorflow for Windows is available only for Python 3.5, and Python 2.7 is not supported even when compiling from source (see e.g. [cmake instructions](https://github.com/tensorflow/tensorflow/tree/r0.12/tensorflow/contrib/cmake)).
Thus, Windows support relies critically on the support for Python3 (see PR #461).

arijit17:
When can we expect Magenta to work on Windows 10?

jsleep:
@arijit17 That would be once all of Magenta's source code has been translated to Python 3, which is probably not at the top of the priority list.

gbernal:
Ah.. I came here looking for the answer to this questions.. Is everyone running some form of Linux?

icoxfog417:
You can use [bash on Windows](https://msdn.microsoft.com/ja-jp/commandline/wsl/about). My platform is Windows 10 and executes the Magenta by bash on Windows. 
I hope the short instruction on my repository helps you.

[icoxfog417/magenta_session#install](https://github.com/icoxfog417/magenta_session#install)

gbernal:
thanks for the advice. I'm still having some issues. I'm now having this issue https://stackoverflow.com/questions/42013316/after-building-tensorflow-from-source-seeing-libcudart-so-and-libcudnn-errors 
Do you install windows cuDNN  or linux cuDNN?

icoxfog417:
@gbernal Now [bash on Windows is not supported GPU](https://github.com/Microsoft/BashOnWindows/issues/829).  We have to wait Magenta supports Python3 or bash on Windows supports GPU!

AlexandreJup:
Hello ! 
Any news on running magenta in windows? (not using bash on Windows)

dubreuia:
@AlexandreJup I'm using Magenta in Windows (with cmd, and PyCharm), no problem for now, so it seems the support for Python 3.x has been fixed. I haven't tested all the models and training for now, but it looks OK. I'll report back if I see anything that doesn't work.

Some things to know:

- You will need Python 3.5.x (I'm using 3.5.6), because `tensorflow-gpu` requires it (not sure if necessary if you only need `tensorflow`)
- You will need CUDA 10.0 (not 10.1), same reason
- I'm using CudRNN 7.6 (not sure if there is a specific version to have), same reason

[...]

PR #461, opened 2016-12-15
WIP adding support for python3

Mistobaan:
WIP on porting magenta to python3, posting to get early comments

bazel test --test_output=streamed --verbose_failures //magenta/... --host_force_python=py3

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -1,9 +1,8 @@
-new_http_archive(
+new_git_repository(
   name = "pretty_midi",
   build_file = "pretty_midi.BUILD",
-  url = "https://github.com/craffel/pretty-midi/archive/0.2.6.tar.gz",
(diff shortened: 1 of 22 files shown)


WORKSPACE
```diff
@@ -1,9 +1,8 @@
-new_http_archive(
+new_git_repository(
```
Mistobaan: there is a problem with the packaging of the remote repo. It contains an invalid python file.

adarob: @Mistobaan does this change actually work? We originally were using new_http_archive because we needed the strip_prefix option to remove the outer 'pretty_midi' directory.

dubreuia:
Isn't that fixed? It looks like it is working but I haven't tested all the models yet. See #429

cghawthorne:
Yes, we now fully support python 3. This PR is out of date, so I'll close it.

PR #506, opened 2017-01-28
Fix up to work with TF v1.0 API.

darrengarvey:
rnn_cell has moved to tensorflow.contrib.rnn. This change was made in
v0.12 [1].

[1] See https://github.com/tensorflow/tensorflow/commit/37fbebdd6c3c8f274896cc36e6feb5b7e2097a59

### DIFF
--- a/magenta/models/rl_tuner/rl_tuner_ops.py
+++ b/magenta/models/rl_tuner/rl_tuner_ops.py
@@ -297,17 +297,14 @@ def make_rnn_cell(rnn_layer_sizes, state_is_tuple=False):
         and cell matrix as a state instead of a concatenated matrix.
 
   Returns:
-      A tf.contrib.rnn.rnn_cell.MultiRNNCell based on the given
-      hyperparameters.
+      A tf.contrib.rnn.MultiRNNCell based on the given hyperparameters.
   """
   cells = []
   for num_units in rnn_layer_sizes:
-    cell = tf.contrib.rnn.rnn_cell.LSTMCell(
-        num_units, state_is_tuple=state_is_tuple)
+    cell = tf.contrib.rnn.LSTMCell(num_units, state_is_tuple=state_is_tuple)
     cells.append(cell)
 
-  cell = tf.contrib.rnn.rnn_cell.MultiRNNCell(
-      cells, state_is_tuple=state_is_tuple)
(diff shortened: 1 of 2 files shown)


magenta/models/rl_tuner/rl_tuner_ops.py
```diff
@@ -21,6 +21,7 @@
 
 import numpy as np
 import tensorflow as tf
+from tensorflow.contrib.rnn import rnn_cell
```
cghawthorne: to be consistent with the rest of our code, we'd prefer to just have one `import tensorflow as tf`, and then have the below references be things like `tf.contrib.rnn.LSTMCell`.

darrengarvey: I should have noticed that was your preferred style. I've made the change: now, do you prefer force-pushing the fixup to keep the PR clean or add an additional commit on top?

darrengarvey: I've pushed as an extra commit, but can squash down if that's how you do it.

magenta/models/rl_tuner/rl_tuner_ops.py
```diff
@@ -297,15 +297,17 @@ def make_rnn_cell(rnn_layer_sizes, state_is_tuple=False):
...
 
   Returns:
-      A tf.nn.rnn_cell.MultiRNNCell based on the given hyperparameters.
+      A tf.contrib.rnn.rnn_cell.MultiRNNCell based on the given
+      hyperparameters.
   """
   cells = []
   for num_units in rnn_layer_sizes:
-    cell = tf.nn.rnn_cell.LSTMCell(
+    cell = tf.contrib.rnn.rnn_cell.LSTMCell(
```
cghawthorne: Need to remove the extra rnn_cell so it's just tf.contrib.rnn.LSTMCell

magenta/models/rl_tuner/rl_tuner_ops.py
cghawthorne: Extra commits are fine. We'll squash before we merge.

darrengarvey:
Removed the superfluous references to `rnn_cell`. I hadn't noticed that ugliness became unnecessary.

I also fixed up a doc reference to `RnnCell`, which AFAICT should have been `RNNCell`.

PR #510, opened 2017-02-01
Adding ranges for the softmax temperature control.

korymath:
The temperature_control_number on magenta_midi represents a temperature, but the range is not immediately clear. By adding details in the flag definition users can more easily understand how to tune the temperature.

### DIFF
--- a/magenta/interfaces/midi/magenta_midi.py
+++ b/magenta/interfaces/midi/magenta_midi.py
@@ -90,8 +90,8 @@ tf.app.flags.DEFINE_integer(
     'temperature_control_number',
     None,
     'The control change number to use for controlling softmax temperature.'
-    'This should be an integer between 0 and 127 corresponding to a softmax '
-    'temperature between 0.1 and 2.')
+    'The value of control changes with this number will be used to set the '
+    'temperature in a linear range between 0.1 and 2.') 
 tf.app.flags.DEFINE_boolean(
     'allow_overlap',
     False,
--- a/magenta/interfaces/midi/magenta_midi.py
+++ b/magenta/interfaces/midi/magenta_midi.py
@@ -89,7 +89,7 @@ tf.app.flags.DEFINE_integer(
 tf.app.flags.DEFINE_integer(
     'temperature_control_number',
     None,
-    'The control change number to use for controlling softmax temperature.')
+    'The control change number to use for controlling softmax temperature. This should be an integer between 0 and 127 corresponding to a softmax temperature between 0.1 and 2.')
 tf.app.flags.DEFINE_boolean(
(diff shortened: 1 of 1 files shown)


korymath:
I signed it!

magenta/interfaces/midi/magenta_midi.py
```diff
@@ -89,7 +89,7 @@
 tf.app.flags.DEFINE_integer(
     'temperature_control_number',
     None,
-    'The control change number to use for controlling softmax temperature.')
+    'The control change number to use for controlling softmax temperature. This should be an integer between 0 and 127 corresponding to a softmax temperature between 0.1 and 2.')
```
adarob: Thanks. Could you please wrap this so that the line has a max length of 80 chars?

magenta/interfaces/midi/magenta_midi.py
```diff
@@ -89,7 +89,9 @@
 tf.app.flags.DEFINE_integer(
     'temperature_control_number',
     None,
-    'The control change number to use for controlling softmax temperature.')
+    'The control change number to use for controlling softmax temperature.'
+    'This should be an integer between 0 and 127 corresponding to a softmax '
+    'temperature between 0.1 and 2.')
```
adarob: Re-reading this, I realize that it is actual not correct. It should really say something like 'The value of control changes using this control number will be used to set the temperature in a range between 0.1 and 2.'

korymath: My hope was that I could add that while the temperature_control_number is between 0 and 127, this value is mapped to between 0.1 and 2. 

I was trying to follow the convention of the file, which lists for each of the flags:

"The control change number to use for..." or "The control change number to use with..." 

What about:

The control change number to use for controlling softmax temperature. The value of this control change number will be used to set the temperature in a linear range between 0.1 and 2. 

I guess that it is clear once you trace how it is being used here: https://github.com/tensorflow/magenta/blob/824ae077abcecc58c568b1a770c6c915dbca616b/magenta/interfaces/midi/midi_interaction.py#L103

adarob: I agree that it's good to include the range. However, I want to make it clear that the flag value doesn't directly determine the temperature. Instead it specifies which control change to monitor whose value will be mapped to the range. How about:

The control change number to use for controlling softmax temperature. The value of control changes with this number will be used to set the temperature in a linear range between 0.1 and 2.

korymath: Love it. That makes much more sense and should help temperature fiddlers, like myself, to work with this interface. I will propose that change ASAP.

PR #513, opened 2017-02-08
Tiny documentation fix

notjosh:
In absent-minded "how does this work?" mode, I totally missed the missing `\`, instead delving way deeper to find a problem. Hopefully this saves someone else 5 minutes in future :)


notjosh:
I signed it!

PR #520, opened 2017-02-13
Remove dangling 'r' variable decleration in version.py

brannondorsey:


brannondorsey:
I signed it!

cghawthorne:
In this case, the `r` indicates that the following string is a raw string: https://docs.python.org/2.7/reference/lexical_analysis.html#string-literals

We tend to do this for our docstrings in case the string contains code or command line snippets.

brannondorsey:
Ah, thank you @cghawthorne, I haven't used docstrings much and didn't know that. Sorry for adding noise.

PR #539, opened 2017-02-28
fix bug in log-likelihood computation

iansimon:


PeterMitrano:
can confirm, works for me. tested with [this code](https://gist.github.com/PeterMitrano/5575de94ae96edbc50e99dca84870e2e)

PR #572, opened 2017-03-21
Support output files in current directory without './'

astromme:
convert_dir_to_note_sequences.py doesn't support specifying an output file in the current directory unless the user prefixes it with './' or uses the full path. This is because the script tries to create the output dir '' (empty string) which fails. This fixes the problem by not creating an output_dir if the output_file doesn't contain a directory component. 

Example command that is fixed by this change:
$ convert_dir_to_note_sequences --input_dir=lmd_matched --output_file output.tfrecord --recursive

### DIFF
--- a/magenta/scripts/convert_dir_to_note_sequences.py
+++ b/magenta/scripts/convert_dir_to_note_sequences.py
@@ -168,9 +168,10 @@ def main(unused_argv):
     return
 
   input_dir = os.path.expanduser(FLAGS.input_dir)
+  output_dir = os.path.dirname(output_file)
   output_file = os.path.expanduser(FLAGS.output_file)
 
-  if not os.path.exists(os.path.dirname(output_file)):
+  if output_dir != '' and not os.path.exists(output_dir):
     os.makedirs(os.path.dirname(output_file))
 
   with note_sequence_io.NoteSequenceRecordWriter(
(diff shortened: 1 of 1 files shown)


magenta/scripts/convert_dir_to_note_sequences.py
```diff
@@ -168,9 +168,10 @@ def main(unused_argv):
     return
 
   input_dir = os.path.expanduser(FLAGS.input_dir)
+  output_dir = os.path.dirname(output_file)
   output_file = os.path.expanduser(FLAGS.output_file)
 
-  if not os.path.exists(os.path.dirname(output_file)):
+  if output_dir != '' and not os.path.exists(output_dir):
```
cghawthorne: We're trying to switch to using tf.gfile when we can for internal compatibility. It doesn't complain when the output directory already exists, so I think you could do:

```
if output_dir:
  tf.gfile.MakeDirs(output_dir)
```

astromme: Thanks. Tested tf.gfile.MakeDirs and it works like you say. Updated my code to use it.

astromme:
Hey! Also good to hear from you, miss you guys. Can you take another look with the updated code?

PR #576, opened 2017-03-23
Fix processing a sequence without notes

cghawthorne:
Bug occurred only when drop_events_n_seconds_after_last_note was used.


cghawthorne:
Updated to handle non-note events. Can you take another look?

PR #594, opened 2017-04-11
Adding licenses to NSynth BUILD files

jesseengel:


magenta/models/nsynth/BUILD
```diff
@@ -1,3 +1,21 @@
+# Copyright 2016 Google Inc. All Rights Reserved.
```
adarob: These should all be 2017.

PR #601, opened 2017-04-21
NSynth: Support long audio files by chunking them

veqtor:
Save embeddings in NSynth for longer/shorter audio files than sample_length
Also support for loading mp3-files using FFMPEG backend
Use librosa for loading audio files to resample
Resampling can cause inter-sample distortion (samples above/below 1 & -1 ) so normalize was added to make sure no samples produce values outside of one-hot encoding

### DIFF
--- a/magenta/models/nsynth/utils.py
+++ b/magenta/models/nsynth/utils.py
@@ -19,10 +19,8 @@ from __future__ import print_function
 
 import importlib
 
-# internal imports
 import librosa
 import numpy as np
-import scipy.io.wavfile
 import tensorflow as tf
 
 slim = tf.contrib.slim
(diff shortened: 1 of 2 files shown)


veqtor:
I signed it! (CLA)

veqtor:
I signed it! (CLA - For other ssh key)

veqtor:
Perhaps this, since it's a bit different from the original implementation should be moved out to a separate method?

jesseengel:
Sorry for the extremely long delay on this. We had to change a bunch of the code to incorporate the new sampler. If you want to update this for the new code and think it'd be useful I'd be happy to take a look. However, using a TitanX I was able to create encodings and generate from a 3 minute song, so I'm not sure what use cases this would really be addressing.

PR #617, opened 2017-05-04
Added bokeh-based piano roll plotting

douglaseck:


magenta/music/notebook_utils.py
```diff
@@ -16,6 +16,11 @@
 # internal imports
 import IPython
 
+import collections
+import bokeh
```
cghawthorne: We should update magenta/tools/pip/setup.py to have dependencies on pandas and bokeh.

Should also add a dependency in the BUILD file with "# bokeh dep" and "# pandas dep" that we can handle internally.

douglaseck: Done. Thanks!

magenta/music/notebook_utils.py
```diff
@@ -16,6 +16,11 @@
 # internal imports
 import IPython
 
+import collections
```
cghawthorne: Should change import order:

```
import collections

# internal imports

import bokeh
import bokeh.plotting
import IPython
import pandas as pd

from magenta.music import midi_synth
```

magenta/music/notebook_utils.py
```diff
@@ -38,3 +43,72 @@ def play_sequence(sequence,
   array_of_floats = synth(sequence, sample_rate=sample_rate, **synth_args)
   IPython.display.display(IPython.display.Audio(array_of_floats,
                                                 rate=sample_rate))
+
+
+def plot_sequence(sequence,
+                  show_figure=True):
+  """Creates an interactive pianoroll for a tensorflow.magenta.NoteSequence.
```
cghawthorne: Should have a blank line after the one-sentence docstring summary.

douglaseck: Fixed.

PR #628, opened 2017-05-08
Fixed decoder transpose in rl_tuner_ops

herougo:
Each entry in event_list is a list where 0 and 1 are special values, 2 is C, and so on. When transpose_amount is 0 (ie stay in C major) and e is 2 (ie C),we expect 48 since that is the midi value for C. In the original code,

e + INITIAL_MIDI_VALUE - transpose_amount = 2 + 48 - 0 = 50 != 48

As a result, C major melodies are transposed to D major without specifying it.


cghawthorne:
Good catch! Could you write a unit test to verify the fix?

herougo:
Where would the unit test go?

cghawthorne:
I think you should be able to add it to rl_tuner_test.py fairly easily.

PR #630, opened 2017-05-08
Fix pad_amt calculation.

adarob:


magenta/models/shared/events_rnn_model.py
```diff
@@ -143,7 +143,7 @@ def _generate_step(self, event_sequences, inputs, initial_states,
     final_states = []
     loglik = np.empty(num_seqs)
 
-    pad_amt = len(event_sequences) % batch_size
+    pad_amt = -len(event_sequences) % batch_size
```
cghawthorne: Can you add some comments about what's happening here?

cghawthorne: Also, how difficult would it be to add a test for this case?

adarob: Added comments. Unfortunately it's not easy to add a test just for this. We really need a test for this entire module, but I don't think we should block this fix for it.

Issue #556, opened 2017-03-14
Polyphony_rnn_train, Incompatible shapes: [50880] vs. [51520]


MarkYHZhang:
I attempted to run:

`polyphony_rnn_train --run_dir=/home/mark/ML/workplace/chopin/polyphony_rnn/logdir/run1 --sequence_example_file=/home/mark/ML/workplace/chopin/sequence_examples/training_poly_tracks.tfrecord --hparams={'rnn_layer_sizes':[64,64],'batch_size':64} --num_training_steps=20000`

But it terminated with this error:
`INFO:tensorflow:Error reported to Coordinator: <class 'tensorflow.python.framework.errors_impl.InvalidArgumentError'>, Incompatible shapes: [50880] vs. [51520]
         [[Node: mul_6 = Mul[T=DT_FLOAT, _device="/job:localhost/replica:0/task:0/cpu:0"](ToFloat_3, Reshape_4)]]

Caused by op u'mul_6', defined at:
  File "/home/mark/miniconda2/envs/magenta/bin/polyphony_rnn_train", line 11, in <module>
    sys.exit(console_entry_point())
  File "/home/mark/miniconda2/envs/magenta/lib/python2.7/site-packages/magenta/models/polyphony_rnn/polyphony_rnn_train.py", line 97, in console_entry_point
    tf.app.run(main)
  File "/home/mark/miniconda2/envs/magenta/lib/python2.7/site-packages/tensorflow/python/platform/app.py", line 43, in run
    sys.exit(main(sys.argv[:1] + flags_passthrough))
  File "/home/mark/miniconda2/envs/magenta/lib/python2.7/site-packages/magenta/models/polyphony_rnn/polyphony_rnn_train.py", line 78, in main
    mode, config, sequence_example_file)
  File "/home/mark/miniconda2/envs/magenta/lib/python2.7/site-packages/magenta/models/shared/events_rnn_graph.py", line 149, in build_graph
    tf.to_float(tf.equal(labels_flat, no_event_label)) * mask_flat)
  File "/home/mark/miniconda2/envs/magenta/lib/python2.7/site-packages/tensorflow/python/ops/math_ops.py", line 814, in binary_op_wrapper
    return func(x, y, name=name)
  File "/home/mark/miniconda2/envs/magenta/lib/python2.7/site-packages/tensorflow/python/ops/math_ops.py", line 987, in _mul_dispatch
    return gen_math_ops.mul(x, y, name=name)
[...]

PR #551, opened 2017-03-12
Fix  'skip_first_n_losses' bug.

YoshikawaMasashi:
Fixes #525.

PR #639, opened 2017-05-13
Update polyphony_model.py

miyasaca:
For fixing 'skip_first_n_losses' bug. #556


cghawthorne:
The skip_first_n_losses bug should have been fixed by #551. Are you still having problems? Are you using the latest pip package?

PR #650, opened 2017-05-18
Update instructions for magenta_midi to make them more clear.

adarob:


magenta/interfaces/midi/README.md
```diff
@@ -130,24 +130,40 @@ magenta_midi --list_ports
...
+BUNDLE_PATHS=/path/to/bundle1.mag,/path/to/bundle2.mag
+```
 
-You will now start the interface with this command, supplying the location of
-the .mag bundle file and any additional flags required by the interaction (see
-below):
+You may now start the interface with this command:
 
 ```bash
 magenta_midi \
```
cghawthorne: I think it may be helpful to include setting the environment variables here again, just so people don't have to read through all the text above to figure out what they should look like

adarob: It's a bit tricky since the values are going to vary for each person. Is this more clear?

PR #668, opened 2017-05-23
Fast WaveNet generation using queues (NSynth)

pkmital:
Here is an implementation of using queues for the WaveNet decoder in NSynth as described in:
Ramachandran, P., Le Paine, T., Khorrami, P., Babaeizadeh, M., Chang, S., Zhang, Y., … Huang, T. (2017). Fast Generation For Convolutional Autoregressive Models, 1–5.

This should let you encode using the existing NSynth model and then synthesize from any encoding using a much faster method than the current approach.  You can generate a 4 second audio file in a few minutes this way, which isn't terrible.  I can get about 100 samples per second using this method (not at all accurate measurements), which means a 4 second clip @ 16 KHz can be synthesized in about 10 minutes.  You can potentially use this to also explore different encodings from interpolation or encode your own sounds and explore their syntheses with this generation method much more easily than before.

There is no CLI tool I'm afraid but I'm hoping someone else can develop that to make it easier for others!  This just includes a simple python module `magenta.models.nsynth.wavenet.generate` which includes a function `synthesize` showing how to use the `FastGenerationConfig` to load an audio file, encode it, and then synthesize from the encoding.

Lastly, I wasn't familiar with the BUILD system so please let me know if that looks okay.

### DIFF
--- a/magenta/models/nsynth/wavenet/BUILD
+++ b/magenta/models/nsynth/wavenet/BUILD
@@ -31,6 +31,27 @@ py_library(
 )
 
 py_library(
+    name = "fastgen",
+    srcs = ["fastgen.py"],
+    deps = [
+        # tensorflow dep
+    ],
+)
+
+py_library(
+    name = "generate",
+    srcs = [
+        "generate.py",
+    ],
+    deps = [
+        ":configs",
+        "//magenta/models/nsynth:utils",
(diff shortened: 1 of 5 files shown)


pkmital:
I signed it!

pkmital:
I signed it!

PR #669, opened 2017-05-23
Fast WaveNet generation using queues (NSynth) (CLA)

pkmital:
Here is an implementation of using queues for the WaveNet decoder in NSynth as described in:
Ramachandran, P., Le Paine, T., Khorrami, P., Babaeizadeh, M., Chang, S., Zhang, Y., … Huang, T. (2017). Fast Generation For Convolutional Autoregressive Models, 1–5.

This should let you encode using the existing NSynth model and then synthesize from any encoding using a much faster method than the current approach. You can generate a 4 second audio file in a few minutes this way, which isn't terrible. I can get about 100 samples per second using this method (not at all accurate measurements), which means a 4 second clip @ 16 KHz can be synthesized in about 10 minutes. You can potentially use this to also explore different encodings from interpolation or encode your own sounds and explore their syntheses with this generation method much more easily than before.

There is no CLI tool I'm afraid but I'm hoping someone else can develop that to make it easier for others! This just includes a simple python module magenta.models.nsynth.wavenet.generate which includes a function synthesize showing how to use the FastGenerationConfig to load an audio file, encode it, and then synthesize from the encoding.

Lastly, I wasn't familiar with the BUILD system so please let me know if that looks okay.


magenta/models/nsynth/wavenet/generate.py
```diff
@@ -0,0 +1,146 @@
+import tensorflow as tf
+from scipy.io import wavfile
+from magenta.models.nsynth import utils
+from magenta.models.nsynth.wavenet.h512_bo16 import Config, FastGenerationConfig
+import numpy as np
+
+
+def inv_mu_law(x, mu=255.0):
```
jesseengel: These functions should probably be loaded from / added to utils.py. You could just rename them as inv_mu_law_numpy() for example.

magenta/models/nsynth/wavenet/generate.py
```diff
@@ -0,0 +1,146 @@
...
+    out: The decoded data.
+  """
+  x = np.array(x).astype(np.float32)
+  out = (x + 0.5) * 2. / (mu + 1)
+  out = np.sign(out) / mu * ((1 + mu)**np.abs(out) - 1)
+  out = np.where(np.equal(x, 0), x, out)
+  return out
+
+
+def load_audio(wav_file, sample_length=64000):
```
jesseengel: These functions should probably be loaded from / added to utils.py. You could just rename them as inv_mu_law_numpy() for example.

PR #672, opened 2017-05-24
sketch_rnn: allow loading of multiple datasets

dribnet:
A simple extension to allow training on the concatenation of multiple
datasets, as was done in the paper. This is done by passing an array in the existing
hparams.data_set field. Change was made backwards compatible.

Some dataset diagnostics were also added to the output. Sample output below.
```bash
INFO:tensorflow:data_set = ['horse.npz', 'zebra.npz']
...
INFO:tensorflow:Loading data files.
INFO:tensorflow:Loaded 70000/2500/2500 from horse.npz
INFO:tensorflow:Loaded 70000/2500/2500 from zebra.npz
INFO:tensorflow:Dataset combined: 150000 (140000/5000/5000), avg len 60
```

### DIFF
--- a/magenta/models/sketch_rnn/sketch_rnn_train.py
+++ b/magenta/models/sketch_rnn/sketch_rnn_train.py
@@ -102,19 +102,44 @@ def load_dataset(data_dir, model_params, inference_mode=False):
   # normalizes the x and y columns usint the training set.
   # applies same scaling factor to valid and test set.
 
-  data_filepath = os.path.join(data_dir, model_params.data_set)
-  if data_dir.startswith('http://') or data_dir.startswith('https://'):
-    tf.logging.info('Downloading %s', data_filepath)
-    response = requests.get(data_filepath)
-    data = np.load(StringIO(response.content))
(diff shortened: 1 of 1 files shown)


hardmaru:
Hi @dribnet 

Thanks for the modification! I think the multi-dataset idea is great! Before I was just manually concatenating them to create a new .npz file.

I tested your code and there isn't any current issue.  However, we are planning on migrating from our HParams class to the tf.HParams class (https://www.tensorflow.org/versions/master/api_docs/python/tf/contrib/training/HParams) later on, and it seems for tf.HParams, it has trouble setting what was originally a single-valued parameter to a list.

One solution is to sacrifice backwards compatibility, and ask you to also modify "get_default_hparams" so that by default, data_set is a list, and assume that the model_config json files also store this as a list.

After doing so, I'll update the sketch_rnn.zip so that the JSON model_config files also conform to the new standard, and will merge the pull request and update the zip file on http://download.magenta.tensorflow.org/ at the same time.

Let us know what you think later.

PR #678, opened 2017-05-25
moved NIPS and ai-duet into a demos folder

jesseengel:
Not a huge change, but will be helpful to have demos all in one place as we add more (nsynth etc.)


adarob:
I also think we should consider renaming NIPS_2016 to something like ai-jam

jesseengel:
Sounds reasonable. Maybe `ableton-jam`, or `ableton-ai-jam`, to make it more clear? @cghawthorne , should this be inside or outside the main magenta folder? The jam does have a dependency on having the pip package installed, but doesn't need to include anything in the pip package itself.

Issue #674, opened 2017-05-24
MusicXML Time Signature Parsing issues


cghawthorne:
We found several issues with the MusicXML parser while trying to parse this file: http://www.choeurlimours.com/private/partitions/P292_79_150421181329.mxl

The result of parsing this file is that there are many more time signatures in the resulting NoteSequence than in the MusicXML file, and many of them have a numerator of 0.

I did a little bit of investigation, and the issue is in `_fix_time_signature`. The two main problems were:
* If there is a measure that contains nothing but `<forward duration='16' />` without any notes, then this is interpreted as a duration of 0.
* For part of this particular score, it has a time signature of 12/8, but in `_fix_time_signature`, that is interpreted using `Fraction`, which turns it in to 3/2. This means that the measure appears to have a new time signature since it's different from what was in the score.

cghawthorne:
@jsawruk would you have time to look into this?

jsawruk:
I'll investigate it in a few days. Thanks for your initial analysis.

jsawruk:
This is an unusual file. Finale and Sibelius imported the `<forward duration='16' />` by simply omitting the measures, giving a fractured score layout. Dorico inserted a whole measure rest in those measures. My plan is to insert a rest when encountering a `<forward>` in some cases. It might be ok to do in all cases; I need to think about this more.

cghawthorne:
The file was created in MuseScore, and when you open the file with that program, it interprets it as an empty measure.

I think adding a rest to what would otherwise be an empty measure is a reasonable solution.

jsawruk:
It definitely makes sense to add a whole measure rest if the `<forward>` is the only element in that measure. I am just wondering if it makes sense to add rest for any `<forward>` element. I don't think it would make sense to generalize, because the model might learn too many rests. I will just insert rests for the whole measure rest case.

PR #699, opened 2017-05-31
Fix issue #674: MusicXML Time Signature and Whole Measure

jsawruk:
Fix the Time Signature and Whole Measure rest parsing issues in #674. Time Signatures are no longer changed to reduced fractions. `<forward>` tag in a measure with no notes creates a whole measure rest.


magenta/music/musicxml_parser.py
```diff
@@ -544,6 +545,31 @@ def _parse_forward(self, xml_forward):
                * self.state.seconds_per_quarter)
     self.state.time_position += seconds
 
+    # Issue #674 - If the <forward> element is in a measure without
+    # any <note> elements, treat it as if it were a whole measure
+    # rest by inserting a rest of that duration
+    note_count = len(self.xml_measure.findall('./note'))
```
cghawthorne: I would put this section of code before any parsing of the measure happens. That way, you won't have to duplicate the code for adding the duration of the note to self.duration because it will happen as a part of normal note parsing. In order to make the timing add up, you'll probably also need to remove the <forward> element before parsing happens.

magenta/music/musicxml_parser.py
```diff
@@ -552,43 +578,64 @@ def _fix_time_signature(self):
...
       # in this measure (no time signature or senza misura).
       # Insert the fractional time signature as the time signature
       # for this measure
-
       self.time_signature = TimeSignature(self.state)
       self.time_signature.numerator = fractional_time_signature.numerator
       self.time_signature.denominator = fractional_time_signature.denominator
       self.state.time_signature = self.time_signature
     else:
+      #print self.state.time_signature, self.time_signature
```
cghawthorne: remove?

magenta/music/musicxml_parser.py
```diff
@@ -552,43 +578,64 @@ def _fix_time_signature(self):
...
       # If no global time signature yet, set it to the
       # time signature in this measure
       if self.state.time_signature is None:
         self.state.time_signature = self.time_signature
```
cghawthorne: Given that you now assign time_signature to self.state.time_signature whenever you assign to self.time_signature, is this code still relevant?

magenta/music/musicxml_parser.py
jsawruk: Removed.

magenta/music/musicxml_parser.py
jsawruk: Good catch.

magenta/music/musicxml_parser.py
jsawruk: Good idea. That simplifies the code. I moved this code into the Part class, into a new method called `_repair_empty_measure`. This method is immediately called before a measure is parsed.

magenta/music/musicxml_parser.py
```diff
@@ -416,9 +416,42 @@ def _parse(self, xml_part, score_parts):
...
+    the time cursor, remove the <forward> element and replace
+    with a whole measure rest of the same duration.
+    """
+    # Issue #674 - If the <forward> element is in a measure without
+    # any <note> elements, treat it as if it were a whole measure
+    # rest by inserting a rest of that duration
+    forward_count = len(measure.findall('forward'))
+    note_count = len(measure.findall('note'))
+    if note_count == 0 and forward_count == 1:
+
```
cghawthorne: Remove empty line

magenta/music/musicxml_parser.py
```diff
@@ -552,43 +586,59 @@ def _fix_time_signature(self):
...
       self.time_signature = TimeSignature(self.state)
       self.time_signature.numerator = fractional_time_signature.numerator
       self.time_signature.denominator = fractional_time_signature.denominator
       self.state.time_signature = self.time_signature
     else:
-      # If no global time signature yet, set it to the
-      # time signature in this measure
-      if self.state.time_signature is None:
-        self.state.time_signature = self.time_signature
+
```
cghawthorne: remove empty line

magenta/music/musicxml_parser.py
```diff
@@ -552,43 +586,59 @@ def _fix_time_signature(self):
...
+        if new_time_sig_fraction == fractional_time_signature:
+          new_time_signature.numerator = fractional_time_signature.numerator
+          new_time_signature.denominator = fractional_time_signature.denominator
 
       # Insert a new time signature only if it does not equal the global
       # time signature.
-      if new_time_signature != self.state.time_signature:
+      if (self.time_signature is None
+          and fractional_time_signature != fractional_state_time_signature
```
cghawthorne: please add parenthesis to make the ordering of these boolean evaluations clear.

magenta/music/musicxml_parser.py
```diff
@@ -552,43 +586,59 @@ def _fix_time_signature(self):
...
         new_time_signature = TimeSignature(self.state)
-        new_time_signature.numerator = fractional_time_signature.numerator
-        new_time_signature.denominator = fractional_time_signature.denominator
+        new_time_signature.numerator = numerator
+        new_time_signature.denominator = denominator
+
+        new_time_sig_fraction = Fraction(new_time_signature.numerator,
+                                         new_time_signature.denominator)
+
+        if new_time_sig_fraction == fractional_time_signature:
```
cghawthorne: This section is to preserve the original form of the time signature, right? If that's the case, should it use `numerator` and `denominator` rather than `fractional_time_signature.numerator` and `fractional_time_signature.denominator`?

magenta/music/musicxml_parser.py
jsawruk: Done.

magenta/music/musicxml_parser.py
jsawruk: Done.

magenta/music/musicxml_parser.py
jsawruk: I added parentheses like this:
`and (fractional_time_signature != fractional_state_time_signature)`

magenta/music/musicxml_parser.py
jsawruk: Good catch. I think I just forgot to change this when rewriting the code. I wasn't sure your suggestion would work, but I ran the tests and they all pass. Thanks!

magenta/music/musicxml_parser.py
```diff
@@ -627,8 +625,10 @@ def _fix_time_signature(self):
         new_time_signature.numerator = numerator
         new_time_signature.denominator = denominator
 
-        new_time_sig_fraction = Fraction(new_time_signature.numerator,
-                                         new_time_signature.denominator)
+        #new_time_sig_fraction = Fraction(new_time_signature.numerator,
```
cghawthorne: remove?

magenta/music/musicxml_parser.py
```diff
@@ -637,7 +637,7 @@ def _fix_time_signature(self):
       # Insert a new time signature only if it does not equal the global
       # time signature.
       if (self.time_signature is None
-          and fractional_time_signature != fractional_state_time_signature
+          and (fractional_time_signature != fractional_state_time_signature)
```
cghawthorne: The order of boolean evaluation still isn't as clear as i would like it to be. `and` is has higher binding precedence than `or`, so I think something like this would be more clear:

```
if (pickup_measure or (
    self.time_signature is None and (
      fractional_time_signature != fractional_state_time_signature)):
  ...
```

magenta/music/musicxml_parser_test.py
```diff
@@ -206,6 +227,8 @@ def checkmusicxmlandsequence(self, musicxml, sequence_proto):
             * musicxml_note.note_duration.duration_float(),
             delta=1)
 
+    # Test whole measure rest.
```
cghawthorne: what does this comment apply to?

jsawruk: Not sure, so I removed it.

jsawruk:
I'm ok with these syntax changes.

PR #714, opened 2017-06-08
Add ai-jam-js demo.

adarob:


adarob:
Removed .DS_Store files. We could decide to move the demos later on. For now, we have them outside of the magenta directory.

PR #717, opened 2017-06-08
Fix quantization error during generation resulting in crashes.

adarob:


adarob:
I didn't mean to do it in this PR, but I am about to push a new version, so might as well.

zhou13:
Will this PR make quantization less accurate?  It seems that it reduces the duration of all the notes by one step.

iansimon:
This shouldn't affect the duration of individual notes, but the start/end times of generated sequences.  Are you observing any difference in note durations?

zhou13:
Oh my bad.

PR #724, opened 2017-06-09
Clarifying comments and minor code clean-up in Sketch-RNN

colinmorris:
* explanation of stroke-3 and stroke-5 formats

* clarification of some internal model details

* simplify/remove some circuitous/redundant code

Mostly cleaned up versions of comments I left for myself as I was trying to understand the source. (e.g. it's really hard to construct a google search for a definition of MDN that isn't "Mozilla developer network"). Hoping they might be useful for others.

There should be no functional changes.

### DIFF
--- a/magenta/models/sketch_rnn/model.py
+++ b/magenta/models/sketch_rnn/model.py
@@ -147,12 +147,9 @@ class Model(object):
     else:
       assert False, 'please choose a respectable cell'
 
-    use_recurrent_dropout = False
-    if self.hps.use_recurrent_dropout == 1:
-      use_recurrent_dropout = True
-
-    use_input_dropout = False if self.hps.use_input_dropout == 0 else True
-    use_output_dropout = False if self.hps.use_output_dropout == 0 else True
+    use_recurrent_dropout = self.hps.use_recurrent_dropout
+    use_input_dropout = self.hps.use_input_dropout
+    use_output_dropout = self.hps.use_output_dropout
 
     if hps.dec_model == 'hyper':
       cell = cell_fn(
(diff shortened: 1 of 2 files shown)


colinmorris:
`I signed it!`

magenta/models/sketch_rnn/model.py
```diff
@@ -244,7 +244,9 @@ def build_model(self, hps):
     self.num_mixture = hps.num_mixture
 
     # TODO(deck): Better understand this comment.
```
colinmorris: I'm not sure who deck is, but I hope he would find the new version of this comment more understandable. :)

magenta/models/sketch_rnn/utils.py
```diff
@@ -195,11 +208,15 @@ def __init__(self,
                limit=1000):
     self.batch_size = batch_size  # minibatch size
     self.max_seq_length = max_seq_length  # N_max in sketch-rnn paper
-    self.scale_factor = scale_factor  # divide data by this factor
+    self.scale_factor = scale_factor  # divide offsets by this factor
     self.random_scale_factor = random_scale_factor  # data augmentation method
-    self.limit = limit  # removes large gaps in the data
+    # Removes large gaps in the data. x and y offsets are clamped to have
+    # absolute value no greater than this limit.
```
colinmorris: I don't totally understand this part. I would think "large gaps" means large distances between a "pen up" and "pen down" state, but from my reading of `preprocess()`, it seems like this actually clamps all offsets. Meaning it would also shrink line segments having an x or y component greater than `limit`. Is that intentional? Or maybe in practice this doesn't happen because some earlier preprocessing means there are no line segments with offsets greater than 1000 or other typical `limit` values?

colinmorris:
gentle ping

PR #735, opened 2017-06-14
fix HParams values() method

francescodelduchetto:
fix wrong acces to keyvals attribute of HParams


francescodelduchetto:
I signed it!

PR #737, opened 2017-06-15
hparams fix: update -> parse_json

brendansudol:
TF's native HParams does not have an `update` to add/override hyperparams compared with (now retired) magenta HParams class; this PR updates those references to use `parse_json` (https://www.tensorflow.org/versions/r1.2/api_docs/python/tf/contrib/training/HParams#parse_json)


hardmaru:
Thanks!

brendansudol:
it looks like @francescodelduchetto folded this change into his PR (https://github.com/tensorflow/magenta/pull/735/commits/965ed258cb937b3bb6c13292665b6746d4dcc427), which is now merged, so I think this one is no longer needed!

PR #758, opened 2017-06-27
add Performance RNN model of polyphonic performances with dynamics and expressive timing

iansimon:


iansimon:
Let's do that later, that's another step that I also want to make sure everyone approves of.

Issue #788, opened 2017-07-21
UnboundLocalError in nsynth_generate.py


evantarrh:
Here's the traceback I get when running `nsynth_generate` as suggested in the `nsynth` readme.

```(magenta) lappy-486:other evantarrh$ nsynth_generate --checkpoint_path=/bugs/wavenet-ckpt/model.ckpt-200000 --source_path=/bugs --save_path=/bugs --batch_size=4
Traceback (most recent call last):
  File "/Users/evantarrh/miniconda2/envs/magenta/bin/nsynth_generate", line 11, in <module>
    sys.exit(console_entry_point())
  File "/Users/evantarrh/miniconda2/envs/magenta/lib/python2.7/site-packages/magenta/models/nsynth/wavenet/nsynth_generate.py", line 89, in console_entry_point
    tf.app.run(main)
  File "/Users/evantarrh/miniconda2/envs/magenta/lib/python2.7/site-packages/tensorflow/python/platform/app.py", line 48, in run
    _sys.exit(main(_sys.argv[:1] + flags_passthrough))
  File "/Users/evantarrh/miniconda2/envs/magenta/lib/python2.7/site-packages/magenta/models/nsynth/wavenet/nsynth_generate.py", line 64, in main
    elif source_path.lower().endswith(postfix):
UnboundLocalError: local variable 'postfix' referenced before assignment
```

It should be a quick fix: rather than `postfix`, we should pass in a tuple and check to see if the `source_path` ends with `(".wav", ".npy")`.

PR #789, opened 2017-07-21
Fix UnboundLocalError in nsynth (#788)

evantarrh:
Addresses issue #788—enables users to run `nsynth_generate` on a path that is just a filename.


evantarrh:
I signed it :-)

PR #796, opened 2017-07-25
Add a script and instructions to export a SavedModel binary for style transfer

elibixby:
@keveman

Add a SavedModel binary exporting script.

Also made a slight modification to the code in `model.py` to allow it to support images with unknown dimensions. This is important as the most logical deployment of this model as an API makes no assumptions about image size, instead inferring that from passed JPEGs.

### DIFF
--- a/magenta/models/image_stylization/BUILD
+++ b/magenta/models/image_stylization/BUILD
@@ -100,6 +100,7 @@ py_binary(
     ],
     deps = [
         ":model",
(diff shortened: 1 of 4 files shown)


elibixby:
@shawngit

elibixby:
@fredbertsch ping on this review please. The Kokoro error appears to be unrelated to this CL.

PR #799, opened 2017-07-26
Disable midi_hub_test from running unless directly specified.

### LINKED ISSUE #759: midi_hub_test timeouts when running bazel test
When I run `bazel test //magenta/...` all tests pass except for the midi_hub_test in /magenta/interfaces/midi which times out.

Console output:
`
TIMEOUT: //magenta/interfaces/midi:midi_hub_test (see /home/s/.cache/bazel/_bazel_s/5be385002debe164746de4f288331729/execroot/magenta-dev/bazel-out/local-opt/testlogs/magenta/interfaces/midi/midi_hub_test/test.log).

Executed 40 out of 40 tests: 39 tests pass and 1 fails locally.
`

After checking the test.log file I get this output, which looks like a failed redirection with the exec command. 

`
exec ${PAGER:-/usr/bin/less} "$0" || exit 1
`

As this is the first issue with this test, I'm sure there is some local setup that I'm missing or have wrong. I'm running Ubuntu 16.04 with bazel 0.5.1

cghawthorne:
Until we fix #759.

### DIFF
--- a/magenta/interfaces/midi/BUILD
+++ b/magenta/interfaces/midi/BUILD
@@ -43,6 +43,9 @@ py_test(
     name = "midi_hub_test",
     srcs = ["midi_hub_test.py"],
     srcs_version = "PY2AND3",
+    tags = [
+        "manual",  # TODO(adarob): Remove manual tag once timeout is fixed. #759
+    ],
     deps = [
         ":midi_hub",
         "//magenta",


magenta/interfaces/midi/BUILD
```diff
@@ -43,6 +43,9 @@ py_test(
     name = "midi_hub_test",
     srcs = ["midi_hub_test.py"],
     srcs_version = "PY2AND3",
+    tags = [
+        "manual",  # TODO(adarob): Remove manual tag once timeout is fixed. #759
```
cghawthorne: Chatted offline. Seems to time out even after 900s.

PR #808, opened 2017-08-01
Fix directory creation bug in download_bundle.

douglaseck:
Unless tmp/bundles happens to exist already, this code fails. I guess not too many people are using the Performance_RNN jupyter notebook.


magenta/music/notebook_utils.py
```diff
@@ -132,11 +132,15 @@ def _sequence_to_pandas_dataframe(sequence):
...
 
+  Target directory target_dir will be created if it does not already exist.
+
   Args:
      bundle_name: A string Magenta bundle name to download.
      target_dir: A string local directory in which to write the bundle.
      force_reload: A boolean that when True, reloads the bundle even if present.
   """
+  if not os.path.exists(target_dir):
+    os.makedirs(target_dir)
```
cghawthorne: Should just call:
```
tf.gfile.MakeDirs(target_dir)
```

The gfile version has better internal compatibility and will just return if the directory already exists, so no need to do the os.path.exists check.

douglaseck:
I fixed a bad email address in my .gitconfig.  You can ignore the CLA message.

douglaseck:
PTAL

douglaseck:
Adding comment to trigger cla check

PR #811, opened 2017-08-03
avoid unnecessary copying in beam search when branch_factor = 1

iansimon:


magenta/models/shared/events_rnn_model.py
```diff
@@ -195,8 +195,11 @@ def _generate_branches(self, event_sequences, loglik, branch_factor,
       all_loglik: A 1-D numpy array of event sequence log-likelihoods, with
           length equal to the length of `all_event_sequences`.
     """
-    all_event_sequences = [copy.deepcopy(events)
-                           for events in event_sequences * branch_factor]
+    if branch_factor > 1:
```
cghawthorne: add comment explaining that this is to optimize for speed.

iansimon: Added.

PR #827, opened 2017-08-15
fix division bug in midi_io

iansimon:


magenta/music/midi_io.py
```diff
@@ -107,7 +107,7 @@ def midi_to_sequence_proto(midi_data):
     key_signature = sequence.key_signatures.add()
     key_signature.time = midi_key.time
     key_signature.key = midi_key.key_number % 12
-    midi_mode = midi_key.key_number / 12
+    midi_mode = midi_key.key_number // 12
```
cghawthorne: Please add a unit test to verify this.

iansimon: Added a key signature to one of the test MIDI files to break the test.  Also fixed the same division bug in the test itself.

PR #837, opened 2017-08-22
from six.moves import xrange for Python 3

cclauss:
xrange() is called on line 279.

### DIFF
--- a/magenta/models/nsynth/utils.py
+++ b/magenta/models/nsynth/utils.py
@@ -23,6 +23,7 @@ import os
 # internal imports
 import librosa
 import numpy as np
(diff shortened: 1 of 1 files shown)


cclauss:
kokoro-gcp_ubuntu-py3 failed but the results are only visible to Google employees.  :-(

cghawthorne:
Yeah, sorry about the log visibility thing. It's something we're working on fixing. I think the py3 presubmit failed due to a kokoro server restart. I re-added the label to see if that will reschedule it. If not, we can just override that check and submit anyway, after @jesseengel approves the PR.

PR #838, opened 2017-08-22
Define reload() for Python 3

cclauss:


magenta/models/rl_tuner/rl_tuner.py
```diff
@@ -62,6 +62,11 @@
 def reload_files():
   """Used to reload the imported dependency files (needed for ipynb notebooks).
   """
+  try:
+    reload                        # Python 2
+  except NameError:
+    from importlib import reload  # Python 3
+
   reload(note_rnn_loader)
```
cghawthorne: would it be better to switch to six.reload_module?

cclauss: Good suggestion... Done.

PR #846, opened 2017-08-25
Build universal pip packages now that we support python 3

cghawthorne:


magenta/tools/pip/build_pip_package.sh
```diff
@@ -41,7 +41,7 @@ function main() {
   pushd "${TMPDIR}"
   rm -f MANIFEST
   echo $(date) : "=== Building wheel"
-  python setup.py bdist_wheel $2 # >/dev/null
+  python setup.py bdist_wheel --universal $2 # >/dev/null
```
iansimon: Using protobuf doesn't prevent us from being able to do this, right?

cghawthorne: As far as I can tell, protoc produces code that works in either py2 or py3.

cghawthorne: Also, our kokoro checks verify that our code works in both py2 and py3.

iansimon: I have virtually no understanding of how the protobuf compilation works but it's not pure Python, right?  Or does it generate pure Python and then that's what gets included in the package?

https://packaging.python.org/tutorials/distributing-packages/#id71

cghawthorne: Yup, the compilation happens at pip build time and then we distribute pure python.

iansimon: Got it, sorry to be obstructionist about something I didn't understand :)

PR #849, opened 2017-08-26
Correctly handle transposed key signatures

cghawthorne:


cghawthorne:
@jsawruk if you have time, would appreciate your input on this one too.

jsawruk:
Doesn't [this test](https://github.com/tensorflow/magenta/blob/93f7c72092b3f1532485a7cb1709caca2136b35e/magenta/music/musicxml_parser_test.py#L306) cover this scenario, or is there a new case that needs to be handled? 

The test takes a Clarinet in Bb scale as input and makes sure it is transposed properly to sounding pitch. Look at the [transpose tag](https://github.com/tensorflow/magenta/blob/93f7c72092b3f1532485a7cb1709caca2136b35e/magenta/music/testdata/clarinet_scale.xml#L121) in the test file. Notice that the first pitch in the MusicXML file is G4, but the test compares this to F4 (the sounding pitch), which passes. Specifically, the test compares the flute and clarinet scale files and makes sure they are equal. Their sounding pitches are equal, but the written pitches are different.

The `el_capitan.xml` file is a conductor's score, and contains several transposing instruments of different transposition levels (Bb Clarinet, Eb Alto Sax, Horn in F), as well as a key change. If a use case isn't covered, it can probably be written against this file.

Note that if the `<transpose>` tag is missing or only includes a `<diatonic>` element, then the code will not perform transposition. According to the [MusicXML schema](http://usermanuals.musicxml.com/MusicXML/MusicXML.htm#CT-MusicXML-transpose.htm#kanchor3032), `<chromatic>` is required; `<diatonic>` is optional.

jsawruk:
I see now, the pitches are transposing but the key signatures are not. Will look into this.

jsawruk:
This looks fine for this commit. For a more intelligent approach to transposing, I would recommend the following logic:

- If transposition adds flats to the key signature, continue adding flats until no more flats can be added.
- If transposition adds sharps to the key signature, continue adding sharps until no more sharps can be added.

Examples:

- Key of C major, Trumpet in Bb. Transposing will move the trumpet to the key of D major, which adds two sharps.
- Key of F# major, Trumpet in Bb. Transposing will move the trumpet to the key of Ab major, because there is no key signature for G# major. I checked this logic using Sibelius, and it appears to be the same as the process I described.

PR #854, opened 2017-08-30
Fix import paths for python 3

cghawthorne:


andimarafioti:
The version that is installed with pip3 doesn't include this changes.

Issue #861, opened 2017-09-03
AttributeError: 'bytes' object has no attribute 'encode' at pipeline.py on Python3


icoxfog417:
When running the `run_pipeline_serial` in `pipeline.py` on Python3, it causes `AttributeError: 'bytes' object has no attribute 'encode'`.

The error cause is `encode` after `SerializeToString`. Because [`SerializeToString` returns `bytes` not `str`](https://developers.google.com/protocol-buffers/docs/pythontutorial) without regard to Python2 or 3 (according to the document)..

[pipeline.py](https://github.com/tensorflow/magenta/blob/master/magenta/pipelines/pipeline.py#L381)

```py
     for output in outputs:
        if six.PY3:
          writers[name].write(output.SerializeToString().encode('utf-8'))
        else:
          writers[name].write(output.SerializeToString())
```

So this if statement for PY3 seems to needless.

icoxfog417:
I proposed the fix at #862.

PR #862, opened 2017-09-03
Fix encoding issue in pipeline(#861)

icoxfog417:
Fix the issue in pipeline.

* `SerializeToString` returns bytes, so `encode` is needless.
* At the `pipeline_test.py`, `MockStringProto.SerializeToString` behavior differ according to Python2 or Python3. Because Python2 `str` is byte string but Python3 `str` is not.  So `MockStringProto.SerializeToString` returns bytes in Python2 but str in Python3. `SerializeToString` should return bytes as its definition in protobuf, so Python3 behavior has to be fixed.

### DIFF
--- a/magenta/common/testing_lib.py
+++ b/magenta/common/testing_lib.py
@@ -15,6 +15,7 @@
 
 # internal imports
 import numpy as np
+import six
 
 from google.protobuf import text_format
 
@@ -74,7 +75,11 @@ class MockStringProto(object):
     return MockStringProto(string)
 
   def SerializeToString(self):  # pylint: disable=invalid-name
-    return 'serialized:' + self.string
+    # protobuf's SerializeToString returns binary string
+    if six.PY3:
+      return ('serialized:' + self.string).encode("utf-8")
+    else:
+      return 'serialized:' + self.string
 
   def __eq__(self, other):
     return isinstance(other, MockStringProto) and self.string == other.string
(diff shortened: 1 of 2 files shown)


cghawthorne:
This seems reasonable to me, but I'd appreciate another review. @iansimon could you take a look?

magenta/common/testing_lib.py
```diff
@@ -74,7 +75,11 @@ def FromString(string):  # pylint: disable=invalid-name
     return MockStringProto(string)
 
   def SerializeToString(self):  # pylint: disable=invalid-name
-    return 'serialized:' + self.string
+    # protobuf's SerializeToString returns binary string
```
iansimon: Is it simpler/correct to just use bytes(self.string) here and avoid the Python version check?

icoxfog417: In the Python3, we have to use `bytes` with encoding (like `bytes("Hello", "utf-8")`).  And `bytes` in Python2 can not take encoding arguments.

Python3
```
>>> bytes("Hello world")
TypeError: string argument without an encoding
>>> bytes("Hello world", "utf-8")
b'Hello world'
```

Python2
```
>>> bytes("Hello world")
Hello world
>>> bytes("Hello World", "utf-8")
  File "main.py", line 3, in <module>                                                                                                                                                                                                                                           
    print bytes("Hello World!\n", "utf-8")                                                                                                                                                                                                                                      
TypeError: str() takes at most 1 argument (2 given)
```

So Sad.

iansimon: My lack of Python3 knowledge is showing :)

PR #871, opened 2017-09-07
Change manual install instructions to Python 3.5

SkyrisBactera:
Tensorflow recommends Python 3.5 for all operating systems. On windows, using 2.7 will cause an issue where you can't install magenta through pip because Tensorflow for windows doesn't support 2.7. Also, magenta supports Python 3 anyways, so it shouldn't be an issue.

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,14 +51,14 @@ If the automated script fails for any reason, or you'd prefer to install by
 hand, do the following steps.
 
 First, download the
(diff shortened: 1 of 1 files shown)


SkyrisBactera:
I signed it!

SkyrisBactera:
I realize that the Python 3 support is actually really buggy right now, so I'll close this

PR #874, opened 2017-09-12
Add tempo parsing

cghawthorne:


cghawthorne:
Not sure why the kokoro build failed, and re-adding the label doesn't seem to trigger it to run again, so I'm just going to bypass the check.

PR #875, opened 2017-09-13
Initial support for note durations

cghawthorne:


cghawthorne:
Bypassing py3 checks for now until we figure out what's up with Kokoro.

PR #876, opened 2017-09-13
Update to latest protobuf package

cghawthorne:
Should fix the kokoro errors we've been seeing.


cghawthorne:
Looks like there are still some lingering py3 issues. I'm going to submit this now and investigate those as a separate PR.

Issue #888, opened 2017-09-21
Convert directory serially to save memory


MichaelMa2014:
Different threads in Python share the same CPU core, so using ThreadPool (almost) doesn't speed up the conversion but takes a lot of memory because the resources are not released until all the threads exit. Converting 50000 files, in my experience, takes more than 50GB memory, which makes dealing with big data sets difficult. I will submit a serialized version as a PR in a minute.

mbs6176966:
any efficient solution for this proble? i have increased my RAM from 8G to 32G,but still haven't solve the problem of training big .tfrecord file.

PR #889, opened 2017-09-21
Convert serially to save memory

MichaelMa2014:
[Here is my rationale.](https://github.com/tensorflow/magenta/issues/888) I hope it makes sense~ 😄

### DIFF
--- a/magenta/scripts/convert_dir_to_note_sequences.py
+++ b/magenta/scripts/convert_dir_to_note_sequences.py
@@ -52,7 +52,7 @@ tf.app.flags.DEFINE_string('log', 'INFO',
                            'DEBUG, INFO, WARN, ERROR, or FATAL.')
 
 
-def queue_conversions(root_dir, sub_dir, writer, recursive=False):
+def convert_files(root_dir, sub_dir, writer, recursive=False):
   """Converts files
 
   Args:
@@ -99,7 +99,7 @@ def queue_conversions(root_dir, sub_dir, writer, recursive=False):
             'Unable to find a converter for file %s', full_file_path)
 
   for recurse_sub_dir in recurse_sub_dirs:
-    queue_conversions(root_dir, recurse_sub_dir, writer, recursive)
+    convert_files(root_dir, recurse_sub_dir, writer, recursive)
 
 
 def convert_midi(root_dir, sub_dir, full_file_path):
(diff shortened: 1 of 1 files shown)


cghawthorne:
This seems reasonable to me. I know we've gotten a number of complaints about this script using up too much memory.

Another option would be to not let the queue of items to convert grow beyond a certain size. This would retain whatever benefits there are to multithreading without causing memory issues.

@adarob since you wrote the threaded version, what do you think?

Calvin-Shen:
It seems really solving the problem that memory can not be released in time when generating .tfrecord file from midi files. thanks a lot @MichaelMa2014 @cghawthorne

magenta/scripts/convert_dir_to_note_sequences.py
```diff
@@ -51,14 +52,14 @@
                            'DEBUG, INFO, WARN, ERROR, or FATAL.')
 
 
-def queue_conversions(root_dir, sub_dir, pool, recursive=False):
-  """Queues callables for converting files to NoteSequences via threadpool.
+def queue_conversions(root_dir, sub_dir, writer, recursive=False):
```
adarob: Rename to something like "convert_files"

Calvin-Shen:
it still exists some proble when converting midi files to .tfrecord file. it seems having no memory problem ,but it still killed my progress with not converting complete files.  I don't know whtat causes this problem. @MichaelMa2014 @cghawthorne @adarob

MichaelMa2014:
@Calvin-Shen Please use `sudo dmesg` to check why your OS killed your process.

Calvin-Shen:
I just do with "sudo dmesg" in terminal.and showing as follows:
wechatimg2111508138832_ pic

Calvin-Shen:
I just use your updated code. I hope you help me to solve the problem.  Thank you~ @adarob @MichaelMa2014 @cghawthorne

MichaelMa2014:
Yes it seems this is a memory issue. Could you please try using only part of your dataset and see if the issue still exists? Sometimes one corrupted file can eat up all the memory and it happened to me before.

MichaelMa2014:
@Calvin-Shen Here is an example of a corrupted midi file. It drained all my memory.
```
>>> import pretty_midi as pm
>>> pm.PrettyMIDI("30073.mid")
  File "/home/v-chenxm/.local/lib/python3.5/site-packages/pretty_midi/pretty_midi.py", line 84, in __init__
    ' it is likely corrupt'.format(max_tick)))
ValueError: MIDI file has a largest tick of 4295008119, it is likely corrupt
```

Calvin-Shen:
@MichaelMa2014 ok~thanks for your patient answer！

Issue #858, opened 2017-09-01
File PermissionError at Midi File output (Python3, Win7)


CircuV:
I'm getting the following error on a performance_rnn_generate run (on Magenta Python3 Win7):
Seems to occur when trying to output the generated Midi file. Don't know
why it tris to create a File in Temp before writing to the final destination.

  File "d:\programme\anaconda3\envs\magenta\lib\site-packages\pretty_midi\pretty
_midi.py", line 1374, in write
    mid.save(filename=filename)
  File "d:\programme\anaconda3\envs\magenta\lib\site-packages\mido\midifiles\mid
ifiles.py", line 409, in save
    with io.open(filename, 'wb') as file:
PermissionError: [Errno 13] Permission denied: 'C:\\Users\\Cv\\AppData\\Local\
\Temp\\tmpkh_zo2m3'

Even I run it in as admin

cghawthorne:
Hmm, we're just using tempfile.NamedTemporaryFile to create the temp file. Do you know if that fails in other circumstances? Are the permissions on your local temp dir not set up correctly?

CircuV:
Tnx for your help.
Permission are ok I guess. I can copy files to that directory. 
I have written a small program to check further:

import tempfile
import io
with tempfile.NamedTemporaryFile() as temp_file:
    print(temp_file.name)
    print(type(temp_file.name))
    filename ='C:\\Users\\Cv\\AppData\\Local\\Temp\\tmp6mlkrnn2'
    with io.open(filename, 'wb') as file:
                print(file)
    with io.open(temp_file.name, 'wb') as file:
                print(file)

with output:
C:\Users\Cv\AppData\Local\Temp\tmp0dhl6noj
<class 'str'>
<_io.BufferedWriter name='C:\\Users\\Cv\\AppData\\Local\\Temp\\tmp6mlkrnn2'>
Traceback (most recent call last):
  File "D:\Fmusic\Magenta\tests\tempfile.py", line 14, in <module>
    with io.open(temp_file.name, 'wb') as file:
PermissionError: [Errno 13] Permission denied: 'C:\\Users\\Cv\\AppData\\Local\\Temp\\tmp0dhl6noj'

IO to the Temp dir seems to work if I specify the path manually. 
Looks like the filename has to have those  \\ to work. And tempfile does not create them somehow.
Even it displays the path with \\ in the error message. Strange.

Banus:
[...]

PR #893, opened 2017-09-25
fix tempfile handling for windows (#858)

icoxfog417:
In the windows, re-open `NamedTemporaryFile` causes PermissionError (#858).
To avoid this error, not pass the file name but file-object that is needless to re-open.

I confirmed `midi_io_test.py` passes all test cases on some environments.

* Windows10/Python3
* Mac/Python2

### DIFF
--- a/magenta/music/midi_io.py
+++ b/magenta/music/midi_io.py
@@ -327,5 +327,9 @@ def sequence_proto_to_midi_file(sequence, output_file,
   pretty_midi_object = sequence_proto_to_pretty_midi(
       sequence, drop_events_n_seconds_after_last_note)
   with tempfile.NamedTemporaryFile() as temp_file:
-    pretty_midi_object.write(temp_file.name)
+    pretty_midi_object.write(temp_file)
+    # Before copying the file, flush any contents
+    temp_file.flush()
+    # And back the file position to top (not need for Copy but for certainty)
+    temp_file.seek(0)
     tf.gfile.Copy(temp_file.name, output_file, overwrite=True)
--- a/magenta/music/midi_io_test.py
+++ b/magenta/music/midi_io_test.py
@@ -178,7 +178,7 @@ class MidiIoTest(tf.test.TestCase):
     # write. If that is implemented, this rewrite code should be
     # modified or deleted.
 
-    # When writing to the temp file, use the file object itself instead of 
+    # When writing to the temp file, use the file object itself instead of
     # file.name to avoid the permission error on Windows.
     with tempfile.NamedTemporaryFile(prefix='MidiIoTest') as rewrite_file:
       original_midi = pretty_midi.PrettyMIDI(filename)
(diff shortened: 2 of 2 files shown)


cghawthorne:
Thanks for the fix! Can you add some comments around the lines you changed explaining why they are necessary? I want to make sure we don't accidentally break it in the future if we forget why the change was made.

icoxfog417:
Thank you for the review! It certainly easy to mistake point.
I added the comment to the each test method. And attached the notifications to each line that modified.

Are these enough or too much?

magenta/music/midi_io.py
```diff
@@ -327,5 +327,9 @@ def sequence_proto_to_midi_file(sequence, output_file,
   pretty_midi_object = sequence_proto_to_pretty_midi(
       sequence, drop_events_n_seconds_after_last_note)
   with tempfile.NamedTemporaryFile() as temp_file:
-    pretty_midi_object.write(temp_file.name)
+    pretty_midi_object.write(temp_file)
+    # Before copying the file, flush any contents
+    temp_file.flush()
+    # And back the file position to top
+    temp_file.seek(0)
```
iansimon: Why is the seek necessary?

icoxfog417: Certainly, copy works fine without `seek(0)`.
 But in the Python, to execute `flush` and `seek(0)` before reloading (copying) the file seems to be necessary (see also [ref1](https://stackoverflow.com/questions/22127960/read-from-file-after-write-before-closing) and [ref2](https://stackoverflow.com/questions/3253258/what-is-the-best-way-to-write-the-contents-of-a-stringio-to-a-file)), and it is reasonable.

Following code is a simple code that proves the need of flush & seek.

```py
import shutil
import tempfile


with tempfile.NamedTemporaryFile() as t_1:
    with tempfile.NamedTemporaryFile() as t_2:
        t_2.write("test_test".encode("utf-8"))
        t_2.flush()
        t_2.seek(0)  # without seek(0), print result is []
        shutil.copyfileobj(t_2, t_1, -1)
    t_1.seek(0)  # back the position after copy
    print(t_1.readlines())

```

[`tf.gfile.Copy`](https://www.tensorflow.org/api_docs/python/tf/gfile/Copy) works without `flush` & `seek(0)` (Specifically, it is needed to execute `flush` **or** `seek(0)`). I tried to confirm the implementation of `tf.gfile.Copy`, [but it is a little hard to chase](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/lib/io/file_io.py#L371).

So I took certainty and reasonable implementation. If it is too much, I'm ok to remove the `seek`.

iansimon: It's up to you whether you want to keep the seek(0) or remove it.

icoxfog417: Thank you for the review!
`tf.gfile.Copy` works without `flush` if `seek(0)` is executed. So the behavior of `tf.gfile.Copy` is a little mysterious.

Because of this, I want to keep this implementation if you are ok. I added the comment about it.

PR #918, opened 2017-10-25
Add notap tag to midi_hub_test

cghawthorne:


cghawthorne:
Comment to trigger CLA check.

Issue #904, opened 2017-10-02
Image Stylizing Distortions


justinej:
Hi, I'm having troubles recreating some of the image stylizing results. I want to train their network to transform a single style ([Starry Night](http://hd.wallpaperswide.com/thumbs/the_starry_night-t2.jpg)) onto the ImageNet images. However, I am getting weird distortions that eventually obscure the content of the image. Does anyone have experience with this and have any tips about what to look at? Did anyone successfully train a model right out of the box?

If this is the wrong forum to be posting this kind of question, let me know and I can move it somewhere more appropriate. Also let me know if there's any other useful information you'd like me to include. Thank you!

I believe the problem is coming from the training process, because the stylization works fine when I use the pre-trained Monet checkpoint.  [Here](https://imgur.com/a/kd0ku) is what the transformations on a single image look like, over the progression of the training, with the number of steps in the captions. After 40,000 steps, we get:

![tiger_0_1000](https://user-images.githubusercontent.com/11280032/31062545-e4d51e98-a75d-11e7-824d-52799c7f6db2.png)

System desc: Ubuntu 16.04, CUDA V7.5.17, GeForce GTX 1060 6GB GPU

Only minor modifications to the code before running it:
* removed magenta imports. I'm straight up running the Python files, so that I can work with the local code (for debugging purposes)
* am using a special saver for the training step, to save the checkpoints that I want:
```
  train_saver = tf.train.Saver(
    max_to_keep=1,
    keep_checkpoint_every_n_hours=1.0/120)
```
The commands I am running:
```
python image_stylization_create_dataset.py
  --vgg_checkpoint=$VGG_CHECKPOINT
  --style_files=$STYLE_FOLDER
  --output_file=/tmp/image_stylization/style_images.tfrecord

python image_stylization_train.py
  --train_dir=/tmp/image_stylization/run1/train
  --style_dataset_file=/tmp/image_stylization/style_images.tfrecord
  --num_styles=1
  --train_steps=40000
  --batch_size=10
  --vgg_checkpoint=$VGG_CHECKPOINT
  --imagenet_data_dir=$IMAGENET

python image_stylization_transform.py
  --num_styles=1
  --checkpoint=$MODEL_CHECKPOINT
  --input_image=path/to/tiger/image
  --which_styles="{0:1.0}"
  --output_dir=/tmp/image_stylization/output
```
[...]

PR #921, opened 2017-10-26
Fix a bug in the preprocessing of imagenet data

golnazg:
Imagenet images were normalizing (dividing by 255) twice.


golnazg:
This change probably will resolve issue #904

golnazg:
I signed it!

PR #940, opened 2017-11-01
Bump version for new pip package

cghawthorne:


cghawthorne:
Kokoro error seems to be spurious, so will merge anyway.

PR #947, opened 2017-11-14
Avoid using a temporary file when loading an image

pbaylies:
Reading from temporary files again doesn't work on Windows; fortunately we can use scipy.ndimage.imread() instead, and avoid using a temporary file altogether.

### DIFF
--- a/magenta/models/image_stylization/image_utils.py
+++ b/magenta/models/image_stylization/image_utils.py
@@ -403,14 +403,11 @@ def load_np_image_uint8(image_file):
     A 3-D numpy array of shape [image_size, image_size, 3] and dtype uint8,
     with values in [0, 255].
   """
-  with tempfile.NamedTemporaryFile() as f:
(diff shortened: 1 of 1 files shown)


pbaylies:
@googlebot I signed it!

adarob:
Actually, this copy is needed internally.

pbaylies:
@adarob if that's the case, can you at least test that it works on Windows? Cheers.

adarob:
Ah, sorry I missed that part of the issue. Is this an open issue with Python?

pbaylies:
@adarob There are [some platform-specific differences](https://docs.python.org/3/library/tempfile.html#tempfile.NamedTemporaryFile) with how Windows handles a NamedTemporaryFile. I'll see if I can test this again and get you the specific error I encountered 11 months ago.

adarob:
Thanks @pbaylies. I don't have a Windows machine around to test it on unfortunately (or fortunately?) :)

PR #952, opened 2017-11-19
NSynth multigpu generate

jesseengel:
Allow for running multiple instances of `nsynth_generate` at once, to generate with multiple gpus.


adarob:
LGTM, but is there a matching CL?

jesseengel:
Nope, I'll do that.

PR #953, opened 2017-11-19
added latent constraints notebook

jesseengel:


adarob:
Others have been putting notebooks in magenta-demos and linking them from here, so we should probably stick with that.

PR #959, opened 2017-11-24
include hparams in generate example

stopachka:
For Generate Tracks, we include the following note: 

```markdown
--hparams should be the same hyperparameters used for the training job, although some of them will be ignored, like the batch size.
```

However in our example, we don't actually include the same hparams. 

I suggest we add this, as that note can be easy to miss, and the error that it causes can lead one to some searching


stopachka:
I signed it!

PR #976, opened 2017-12-12
Fix support for internal datasets.

adarob:


cghawthorne:
Looks like this breaks in py2

PR #986, opened 2017-12-19
Add music_vae_train as entry point

MichaelMa2014:
Add `music_vae_train` as entry point so that it can be called directly from command line


magenta/tools/pip/setup.py
```diff
@@ -90,6 +90,7 @@
     'magenta.models.polyphony_rnn.polyphony_rnn_train',
     'magenta.models.rl_tuner.rl_tuner_train',
     'magenta.models.sketch_rnn.sketch_rnn_train',
+    'magenta.models.music_vae.music_vae_train',
```
adarob: Please move this between melody_rnn and nsynth to maintain alphabetical order.

PR #989, opened 2017-12-20
 Add additional MusicVAE documentation and logging flag.

adarob:


magenta/models/music_vae/README.md
```diff
@@ -47,8 +47,18 @@ music_vae_train \
...
+flag for your particular train set and hardware. For example, if the default
+batch size of a config is too large for your GPU, you can reduce the batch size
+and learning rate by setting the flag as follows:
+
+```
+--hparams=batch_size=32,learning_rate=0.0005
+```
+
+Finally, you should also launch an evaluation job (using ``--mode=eval`` with a
+healdout dataset) in order to compute metrics such as accuracy and to avoid
```
cghawthorne: s/healdout/held out/

magenta/models/music_vae/README.md
```diff
@@ -47,8 +47,18 @@ music_vae_train \
 --examples_path=/tmp/music_vae/mel_train_examples.tfrecord
 ```
 
-You should also launch an evaluation job (using ``--mode=eval`` with a heldout
-dataset) in order to compute metrics such as accuracy and to avoid overfitting.
+You will likely need to adjust some of the hyperparamters with the ``--hparams``
```
cghawthorne: shouldn't inline code be a single backtick?

https://guides.github.com/features/mastering-markdown/#syntax

adarob: indeed, they both seem to work, but that saves us some storage costs :)

magenta/models/music_vae/README.md
adarob: Fixed.

Issue #985, opened 2017-12-19
music_vae_train stops after filling buffer


MichaelMa2014:
`music_vae_train` gets stuck after saying "Shuffle buffer filled". I can see that it takes all the memory on a GPU card and also 100% of one CPU core but it has been this way for more than 24 hours now. I am still trying to figure out whether it's a bug and any help is welcome!

adarob:
Hi Michael!

It sounds like there might not be enough examples for it to fill a batch.
If you run it in eval mode, it will report how many training examples it is
able to extract.
Which config are you using and what is your dataset?

-Adam

MarkWuNLP:
Hi Adam,
   @adarob  How much data does music_vae_train need if I use the cat-trio_16bar_big setting? I am not sure whether 0.2 million midi files are enough for this model. 
   Another question: does the music_vae model have some special requirements for the provided data? For example, melody_rnn requires midi files that do not have polyphonic notes. Does music_vae model have similar requirements?
   I am looking for your reply. Thank you in advance.

adarob:
That should be enough to train, assuming most of them meet the
requirements, which is that they have segments with overlapping parts in
all 3 of the following categories: drums (channel 9), melody (program
number 0-31), bass (program number 32-39).
The "melody" and "bass" sections can be polyphonic, but the model will
ignore all but the highest-pitched lines.

-Adam

adarob:
@MichaelMa2014, have you been able to resolve this issue? @MarkWuNLP, are you seeing a similar issue?

MarkWuNLP:
@adarob Thank you for your reply. Michael and I are in the same research group. 

We followed your suggestion and found that the data Michael used is empty after the pre-processing. The dataset we used is Nottingham dataset [https://github.com/jukedeck/nottingham-dataset], which is a polyphonic piano data set. We found that if we put the raw data to music_vae (config = cat-mel2bar-small), it will return 0 melody segment. After we extract its melody by ourselves (extract the piano right hand returned by pretty Midi), the music_vae_train works.

We are checking the data pre-processor of music_vae_train. 

Another question I want to ask is the speed of data pre-process. When we train music_vae with the data set lakh (http://colinraffel.com/projects/lmd/#get), we found that the training speed of music_vae is 1 step per minute.

adarob:
[...]

PR #1001, opened 2018-01-04
Minor updates to MusicVAE.

adarob:
-Set skip_polyphony to false by default for melody models.
-Fix variable scopes.
-Add support for argmax in NADE sampling (faster scheduled sampling training).


adarob:
Fixes #985

PR #1004, opened 2018-01-05
Fix incorrect use CudnnLSTM with scheduled sampling. Simplify related code and add documentation.

adarob:


magenta/models/music_vae/lstm_models.py
```diff
@@ -247,24 +247,39 @@ def _flat_reconstruction_loss(self, flat_x_target, flat_rnn_output):
...
+    """
     initial_state = initial_cell_state_from_embedding(
-        self._dec_cell, z, batch_size, name='decoder/z_to_initial_state')
+        self._dec_cell, z, name='decoder/z_to_initial_state')
 
-    if (isinstance(helper, seq2seq.TrainingHelper) and
-        self._cudnn_dec_lstm):
+    # CudnnLSTM is only a replacement for TrainingHelper since it does not
+    # support sampling.
+    if type(helper) == seq2seq.TrainingHelper and self._cudnn_dec_lstm:  # pylint:disable=unidiomatic-typecheck
```
cghawthorne: would it be helpful to raise an exception if cudnn_dec_lstm = True and scheduled sampling is on?

PR #1017, opened 2018-01-16
Add prefetch to end of data pipeline

cghawthorne:


magenta/models/music_vae/music_vae_train.py
```diff
@@ -138,6 +141,8 @@ def train(train_dir,
       train_dataset = train_dataset.padded_batch(
           config.hparams.batch_size, train_dataset.output_shapes)
```
adarob: can you remove this space just so the dataset creation is a block?

magenta/models/music_vae/music_vae_train.py
```diff
@@ -218,6 +223,9 @@ def evaluate(train_dir,
         dataset
         .padded_batch(config.hparams.batch_size, dataset.output_shapes)
         .take(num_batches))
+
```
adarob: ditto

PR #1026, opened 2018-01-23
provide_batch() got an unexpected keyword argument 'examples_path'

Calvin-Shen:
argument called examples not examples_path in function provide_batch~


cghawthorne:
Thanks for the fix in train_util.py! However, it looks like there's also a change to a README.md that has already been committed in a different PR. Can you remove that change so this PR is just for train_util.py?

Calvin-Shen:
The conflict has been resolved!  @cghawthorne

cghawthorne:
Thanks for the contribution!

PR #1027, opened 2018-01-23
Modifications to data.py to allow for hierarchical models

craffel:


magenta/models/music_vae/data.py
```diff
@@ -227,10 +227,9 @@ def _to_items(self, samples):
     """Implementation that decodes model samples into list of items."""
     pass
 
-  def _maybe_sample_outputs(self, outputs):
+ def _maybe_sample_outputs(self, outputs):
     """If should limit outputs, returns up to limit (randomly if training)."""
-    if (not self.max_tensors_per_item or
```
adarob: We still need this check since we call the method from other places.

magenta/models/music_vae/data.py
```diff
@@ -242,8 +241,11 @@ def _maybe_sample_outputs(self, outputs):
   def to_tensors(self, item):
     """Python method that converts `item` into list of tensors."""
     results = self._to_tensors(item)
-    sampled_results = self._maybe_sample_outputs(zip(*results))
-    return list(zip(*sampled_results)) if sampled_results else ([], [])
+    if self.max_tensors_per_item:
```
adarob: Also make this change in the NoteSequence one. Bonus points if you can use shared code, e.g., by calling this super-method.

craffel: Done.

adarob:
It looks like the tests are failing on this change... http://sponge/9845f68b-88eb-42bf-9dff-35b7c8524cd1

craffel:
It looks like it's failing because `to_tensors` is returning a single value but the code is written to expect two values, is this unrelated?

PR #1054, opened 2018-02-16
Shuffle order of dataset files when training.

adarob:
Fixes internal training bug.

### DIFF
--- a/magenta/models/music_vae/data.py
+++ b/magenta/models/music_vae/data.py
@@ -1039,11 +1039,17 @@ def get_dataset(config, tf_file_reader_class=tf.data.TFRecordDataset,
 
   tf.logging.info('Reading examples from: %s', examples_path)
 
-  filenames = tf.gfile.Glob(examples_path)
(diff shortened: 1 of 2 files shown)


magenta/models/music_vae/data.py
```diff
@@ -1039,10 +1039,11 @@ def get_dataset(config, tf_file_reader_class=tf.data.TFRecordDataset,
 
   tf.logging.info('Reading examples from: %s', examples_path)
 
-  files = tf.data.Dataset.list_files(examples_path)
-  reader = files.apply(
-      tf.contrib.data.parallel_interleave(
-          tf_file_reader_class, cycle_length=num_threads))
+  filenames = tf.gfile.Glob(examples_path)
+  if is_training:
+    random.shuffle(filenames)
```
cghawthorne: would it be better to first do random.shuffle and then pass that to parallel_interleave?

adarob:
Done.

PR #1060, opened 2018-02-21
Upgrade to librosa 0.6.0 api

cghawthorne:
Should fix kokoro failures.

### DIFF
--- a/magenta/models/nsynth/utils.py
+++ b/magenta/models/nsynth/utils.py
@@ -229,7 +229,7 @@ def specgram(audio,
 
     # Magnitudes, scaled 0-1
     if log_mag:
-      mag = (librosa.logamplitude(
+      mag = (librosa.power_to_db(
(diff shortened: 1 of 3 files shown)


magenta/models/nsynth/utils.py
```diff
@@ -229,7 +229,7 @@ def specgram(audio,
 
     # Magnitudes, scaled 0-1
     if log_mag:
-      mag = (librosa.logamplitude(
+      mag = (librosa.power_to_db(
```
iansimon: The documentation for power_to_db doesn't mention a ref_power argument; are you sure that works?

https://librosa.github.io/librosa/generated/librosa.core.power_to_db.html

cghawthorne: Good catch. It was renamed to 'ref' here: https://github.com/librosa/librosa/pull/464

I've updated this code to use 'ref' instead of 'ref_power'.

PR #1067, opened 2018-02-23
Add initial version of MusicVAE deeplearn implementation.

adarob:
@hapticdata
@torinmb


hapticdata:
[for the standalone build](https://github.com/adarob/magenta/blob/c163c91078a3594b97b9a6d3d43747d725eb763c/magenta/models/music_vae/js/package.json#L8) I highly recommend changing `--standalone music-vae` to `--standalone musicVAE`.
Using a hyphen would require the user to access it as `new window["music-vae"].MusicVAE()` as opposed to just `new musicvae.MusicVAE()`.

Also, because [deeplearn](https://github.com/adarob/magenta/blob/c163c91078a3594b97b9a6d3d43747d725eb763c/magenta/models/music_vae/js/package.json#L25) would be a required dependency to install this via npm/yarn it should be a `dependency` and not a `devDependency`

Similarly [http-server](https://github.com/adarob/magenta/blob/c163c91078a3594b97b9a6d3d43747d725eb763c/magenta/models/music_vae/js/package.json#L30) should be a `devDependency` and not a `dependency` because it is only necessary for viewing the example, not as part of a build.

magenta/models/music_vae/js/package.json
```diff
@@ -0,0 +1,32 @@
+{
+  "name": "musicvae",
```
cghawthorne: ```@magenta/model-musicvae```

PR #1073, opened 2018-02-27
Update @magenta/music-vae API with NoteSequence-like I/O.

adarob:
@nsthorat
@torinmb 
@hapticdata
@zischwartz
@currycurry
@themarco


hapticdata:
I would recommend against using underscores in the package name, it is very uncommon for npm, snake case is very uncommon for JS. An npm user would expect `@magenta/musicvae` or `@magenta/music-vae`; lowercase and hyphen separated are the conventions.


Also, it appears that `./data` and `./models` are not committed? I couldn't find `Note`.

I dont know the implementation details of `Note` but I would consider keeping it exclusively for internal use and using pure js arrays instead:

```js
[
  [36, 0], [42, 2], [36, 4], [42, 6],
  [36, 8], [42, 10], [36, 12], [42, 14],
  [36, 16], [36, 24], [36, 28], [42, 30]
]
```

over
```js
[
  new Note(36, 0), new Note(42, 2), new Note(36, 4), new Note(42, 6),
  new Note(36, 8), new Note(42, 10), new Note(36, 12), new Note(42, 14),
  new Note(36, 16), new Note(36, 24), new Note(36, 28), new Note(42, 30)
]
```

adarob:
Thanks for the feedback @hapticdata!

1. Switched to music-vae. For some reason I thought you'd said before the hyphen was a no-no but maybe that was for something different.
2. Note[] is going to be replaced by our NoteSequence proto once @cghawthorne has it checked in (probably tomorrow) anyways.

hapticdata:
Hey Adam,

Here are some examples of the value of the interfaces. If you use the protobuffs internally but accept plain-objects matching the shape, the music-vae interface will look something like this with simple serialized data:

```ts
import { MusicVAE } from '@magenta/music-vae';

const noteSequence: INoteSequence = {
  notes: [
    {pitch: 36, startTime: 0 },
    {pitch: 42, startTime: 2 },
    {pitch: 42, startTime: 6},
    {pitch: 36, startTime: 8},
    {pitch: 42, startTime: 10}, 
    {pitch: 36, startTime: 12}, 
    {pitch: 42, startTime: 14},
    {pitch: 36, startTime: 16}, 
    {pitch: 36, startTime: 24}, 
    {pitch: 36, startTime: 28}, 
    {pitch: 42, startTime: 30}
  ]
};

new MusicVAE(checkPointURL)
  .initialize()
  .then((vae)=> vae.interpolate(noteSequence, 10))
  .then(output=> // do something );
```

If you are requiring them to be instances of NoteSequence and Note it will likely require the user to import 2 libraries (and is also more likely to result in 2 different versions of magenta.js being loaded in):

```ts
import { MusicVAE } from '@magenta/music-vae';
import { NoteSequence, Note } from '@magenta/core';

const noteSequence: NoteSequence = new NoteSequence({
  notes: [
    new Note({pitch: 36, startTime: 0 }), 
    new Note({pitch: 42, startTime: 2 }),
    new Note({pitch: 42, startTime: 6}),
    new Note({pitch: 36, startTime: 8}),
    new Note({pitch: 42, startTime: 10}),
    new Note({pitch: 36, startTime: 12}), 
    new Note({pitch: 42, startTime: 14}),
    new Note({pitch: 36, startTime: 16}),
    new Note({pitch: 36, startTime: 24}), 
    new Note({pitch: 36, startTime: 28}), 
    new Note({pitch: 42, startTime: 30})
  ]
});

new MusicVAE(checkPointURL)
  .initialize()
  .then((vae)=> vae.interpolate(noteSequence, 10))
  .then(output=> // do something );
```

Also if the user is writing vanilla js, they will need 2 script tags and will need to reference 2 global objects and this will result in 2 copies of magenta.js (one loaded globally for the `NoteSequence` and `Note` class and one bundled internally to `musicvae`:

```js
const NoteSequence = window.magenta.NoteSequence;
const Note = window.magenta.Note;
const vae = new musicvae.MusicVAE(checkpointURL);
```

PR #1074, opened 2018-02-28
Switch to dcodeIO/protobuf.js, implement quantizeNoteSequence

cghawthorne:
Also copied deeplearn.js lint and vscode settings.

To regenerate the protobuf code, just do `yarn proto`.


magenta/js/src/sequences.ts
```diff
@@ -0,0 +1,272 @@
...
+   * @throws {BadTimeSignatureException} If the time signature found in
+   *    noteSequence has a 0 numerator or a denominator which is not a power of
+   *    2.
+   * @throws {NegativeTimeException} If a note or chord occurs at a negative
+   *    time.
+   */
+  public static quantizeNoteSequence(noteSequence: NoteSequence,
+      stepsPerQuarter: number): NoteSequence {
+    // Make a copy.
+    const qns = NoteSequence.decode(
```
adarob: Is this really the best way to make a copy??

adarob: If so, can we put this in a helper fn?

cghawthorne: @hapticdata do you have thoughts on a better way to do this? I got this idea from this thread: https://github.com/dcodeIO/protobuf.js/issues/315

PR #1078, opened 2018-02-28
Switch to NoteSequence in musicvae.js

adarob:


magenta/models/music_vae/js/demo/index.ts
```diff
@@ -1,40 +1,77 @@
 // tslint:disable-next-line:max-line-length
-import { MusicVAE, Note, DrumsConverter, DrumRollConverter, MelodyConverter } from '../index';
+import { MusicVAE, Note, DrumsConverter, DrumRollConverter, MelodyConverter, INoteSequence} from '../index';
```
cghawthorne: I think I still may not be fully understanding the NoteSequence vs INoteSequence stuff. Why use INoteSequence here and NoteSequence within the other files?

magenta/models/music_vae/js/data.ts
```diff
@@ -109,25 +95,29 @@ export class DrumsConverter extends DataConverter{
...
-      drumRoll.set(0, note.startStep, -1);
+      drumRoll.set(1, note.quantizedStartStep, this.pitchToClass[note.pitch]);
+      drumRoll.set(0, note.quantizedStartStep, -1);
     });
     return drumRoll.toTensor() as dl.Tensor2D;
   }
 
   toNoteSequence(oh: dl.Tensor2D) {
-    const notes: Note[] = [];
+    const noteSequence = new NoteSequence();
```
cghawthorne: Just an FYI, if we ever switched to using reflection instead of statically generated classes (probably unlikely), then NoteSequence.create() has better compatibility. It probably doesn't matter, but that's why I used .create() in the core library.

https://github.com/dcodeIO/ProtoBuf.js/#pbts-for-typescript

adarob: Done.

adarob:
Ready for review
@cghawthorne 
@hapticdata

PR #1082, opened 2018-03-01
Fix sort bug and add dist/musicvae.js to github.

adarob:
@hapticdata
@torinmb


hapticdata:
Some library authors prefer not to publish their generated bundle file to their repo; but the majority do include it just to keep it as simple and convenient for end-users as possible. 
If you would prefer to put it up on a CDN and link to it in a readme, I think that is fine as well.

nsthorat:
We don't check it in in `dist/` and we only host on CDN / NPM, but I would defer to @hapticdata on this since he's more familiar with the OSS JS world :)

PR #1094, opened 2018-03-07
Add DataConverter spec to include with MusicVAE checkpoints.

adarob:
@torinmb 
@hapticdata

### DIFF
--- a/magenta/models/music_vae/js/dist/musicvae.js
+++ b/magenta/models/music_vae/js/dist/musicvae.js
@@ -24,6 +24,21 @@ var DEFAULT_DRUM_PITCH_CLASSES = [
     [49, 55, 57, 58],
     [51, 52, 53, 59, 82]
 ];
+function converterFromSpec(spec) {
+    if (spec.type === 'MelodyConverter') {
+        return new MelodyConverter(spec.args.numSteps, spec.args.minPitch, spec.args.maxPitch);
+    }
+    else if (spec.type === 'DrumsConverter') {
+        return new DrumsConverter(spec.args.numSteps, spec.args.pitchClasses);
+    }
+    else if (spec.type === 'DrumRollConverter') {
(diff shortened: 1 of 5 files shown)


magenta/models/music_vae/js/dist/musicvae.js
```diff
@@ -24,6 +24,21 @@ var DEFAULT_DRUM_PITCH_CLASSES = [
     [49, 55, 57, 58],
     [51, 52, 53, 59, 82]
 ];
+function converterFromSpec(spec) {
```
cghawthorne: why not use the one from data.ts?

adarob:
@hrldcpr

magenta/models/music_vae/js/dist/musicvae.js
```diff
@@ -387,8 +401,18 @@ var Decoder = (function () {
...
         this.checkpointURL = checkpointURL;
-        this.dataConverter = dataConverter;
+        if (dataConverter) {
+            this.dataConverter = dataConverter;
+        }
+        else {
+            fetch(checkpointURL + '/converter.json')
+                .then(function (response) { return response.json(); })
+                .then(function (converterSpec) {
+                _this.dataConverter = data.converterFromSpec(converterSpec);
```
cghawthorne: Rather than _this, I think you can use .bind: https://stackoverflow.com/questions/16435167/binding-a-promise-handler-function-to-an-object

hrldcpr: You can also use an arrow function, which leaves `this` alone. (Unless you're avoiding arrows in this project.)

magenta/models/music_vae/js/dist/musicvae.js
adarob: this is the bundle

magenta/models/music_vae/js/dist/musicvae.js
adarob: this is the bundle

PR #1100, opened 2018-03-10
use version of tornado to avoid weird kokoro issue

iansimon:


iansimon:
Just testing whether or not requiring tornado version < 5.0 avoids the kokoro failure.

cghawthorne:
Since the problem is only for our kokoro env, what about just modifying test.sh?

PR #1107, opened 2018-03-14
from six.moves import xrange

cclauss:


cclauss:
@adarob The tests do not pass but only insiders can see why.  Can you please let me know what went wrong?

cghawthorne:
As far as I can tell, the problem was with our test infra, not the tests themselves. Should be fine to merge.

PR #1108, opened 2018-03-14
Add temperature to sampling.

adarob:
@hapticdata


hapticdata:
all of this looks good except I don't think its a good idea to use `"deeplearn": "next"` this is will continue to always pull in there latest unstable version.

adarob:
We have to use it for now because dl.multinomial is not in any other
version. I will switch it once the new version is released.

PR #1109, opened 2018-03-14
OMR open source initial commit

ringw:


ringw:
Will work on a new magenta-omr repo instead.

PR #1117, opened 2018-03-20
Add Python3 support to midi_hub and fix tests.

adarob:


magenta/interfaces/midi/midi_hub_test.py
```diff
@@ -345,7 +345,7 @@ def testStartCapture_IsDrum(self):
     captor = self.midi_hub.start_capture(120, start_time)
 
     # Channels are 0-indexed in mido.
```
cghawthorne: is this comment still true?

adarob: Yes, I just thought the one-indexed drum channel was 9 when it is actually 10.

PR #1124, opened 2018-03-22
Add support for single-instrument hierarchical/conductor models in MusicVAE.js.

adarob:


magenta/models/music_vae/js/demo/index.ts
```diff
@@ -180,10 +183,56 @@ async function runMel(){
   console.log(dl.memory());
 }
 
+function concatNoteSequences(
```
cghawthorne: We should just port sequences_lib.concatenate_sequences and put it in core.

magenta/models/music_vae/js/model.ts
```diff
@@ -184,10 +124,113 @@ class Encoder {
   }
 }
 
+/**
+ * A hierarchical encoder that uses the outputs from each level as the inputs
+ * to the subsequent level.
+ */
+class HierarhicalEncoder extends Encoder {
```
cghawthorne: spelling: HierarchicalEncoder

magenta/models/music_vae/js/package.json
```diff
@@ -28,8 +30,10 @@
   },
   "dependencies": {
     "@magenta/core": "^0.0.9",
+    "@types/clone": "^0.1.30",
```
cghawthorne: Do you also need a dependency directly on 'clone'?

magenta/models/music_vae/js/model.ts
adarob: Done

magenta/models/music_vae/js/package.json
adarob: Done.

magenta/models/music_vae/js/demo/index.ts
adarob: Agreed, but I don't want to block this since that is a more complicated version of the function and will need tests. Added a TODO for now.

PR #1125, opened 2018-03-24
melody_rnn javascript

cghawthorne:
data.ts is a copy for now until we figure out how to best handle deeplearn.js in multiple modules.


magenta/js/music_rnn/demo/index.ts
```diff
@@ -0,0 +1,114 @@
...
+ * limitations under the License.
+ */
+
+import * as magenta from '@magenta/core';
+import NoteSequence = magenta.NoteSequence;
+import INoteSequence = magenta.INoteSequence;
+import { MelodyRnn } from '../src/index';
+import * as dl from 'deeplearn';
+
+const MELODY_NS = NoteSequence.create({
```
adarob: This would be a lot cleaner with INoteSequence since you can just pass it in arbitrary objects instead of having to do NoteSequence.blah.create

magenta/js/music_rnn/dist/.gitignore
```diff
@@ -0,0 +1,2 @@
+*
```
adarob: Can't you just have a single .gitignore in the main dir that includes dist?

magenta/js/music_rnn/package.json
```diff
@@ -0,0 +1,37 @@
+{
+  "name": "@magenta/music-rnn",
+  "version": "0.0.1",
+  "description": "",
```
adarob: Add description and unpkg, jsdelivr.

magenta/js/music_rnn/src/model.ts
```diff
@@ -0,0 +1,161 @@
...
+    this.lstmBias1.dispose();
+    this.lstmKernel2.dispose();
+    this.lstmBias2.dispose();
+    this.lstmFcW.dispose();
+    this.lstmFcB.dispose();
+    this.initialized = false;
+  }
+
+  async continueSequence(sequence: magenta.INoteSequence, steps: number,
+    temperature?: number): Promise<magenta.INoteSequence> {
```
adarob: fix indent

magenta/js/music_rnn/src/model.ts
```diff
@@ -0,0 +1,161 @@
...
+    this.lstmKernel1.dispose();
+    this.lstmBias1.dispose();
+    this.lstmKernel2.dispose();
+    this.lstmBias2.dispose();
+    this.lstmFcW.dispose();
+    this.lstmFcB.dispose();
+    this.initialized = false;
+  }
+
+  async continueSequence(sequence: magenta.INoteSequence, steps: number,
```
adarob: Add doc to public methods.

magenta/js/music_rnn/src/model.ts
```diff
@@ -0,0 +1,161 @@
...
+   * Loads variables from the checkpoint and instantiates the `Encoder` and
+   * `Decoder`.
+   */
+  async initialize() {
+    const reader = new dl.CheckpointLoader(this.checkpointURL);
+    const vars = await reader.getAllVariables();
+
+    // Model variables.
+    // tslint:disable:max-line-length
+    this.lstmKernel1 = vars['RNN/MultiRNNCell/Cell0/BasicLSTMCell/Linear/Matrix'] as dl.Tensor2D;
```
adarob: This looks to be the old naming format. We should use the new one and update the checkpoints as needed.

magenta/js/music_rnn/src/model.ts
```diff
@@ -0,0 +1,161 @@
...
+    this.lstmBias1 = vars['RNN/MultiRNNCell/Cell0/BasicLSTMCell/Linear/Bias'] as dl.Tensor1D;
```
adarob: Let's support arbitrary number of layers. You can see how I do this in MusicVAE.

magenta/js/music_rnn/src/model.ts
```diff
@@ -0,0 +1,161 @@
...
+          nextInput = inputs.slice([
+            i, 0], [1, outputSize]).as2D(1, outputSize);
+        } else {
+          const logits = h[1].matMul(this.lstmFcW).add(this.lstmFcB);
+          const sampledOutput = (
+            temperature ?
+            dl.multinomial(logits.div(dl.scalar(temperature)), 1).as1D():
+            logits.argMax(1).as1D());
+          nextInput = dl.oneHot(sampledOutput, outputSize).toFloat();
+          samples.push(nextInput.as1D());
```
adarob: Save the samples as bool or just the labels to reduce data sync time.

magenta/js/music_rnn/src/model.ts
cghawthorne: Done

magenta/js/music_rnn/src/model.ts
cghawthorne: Done

magenta/js/music_rnn/package.json
adarob: Oh, sorry if I missed that before.

magenta/js/core/src/sequences.ts
```diff
@@ -157,11 +171,10 @@ export class Sequences {
    * @throws {NegativeTimeException} If a note or chord occurs at a negative
    *    time.
    */
-  public static quantizeNoteSequence(noteSequence: NoteSequence,
+  public static quantizeNoteSequence(noteSequence: INoteSequence,
     stepsPerQuarter: number): NoteSequence {
     // Make a copy.
-    const qns = NoteSequence.fromObject(
-      NoteSequence.toObject(noteSequence));
+    const qns = NoteSequence.fromObject(noteSequence);
```
adarob: This actually isn't a deep copy, which is why I was using clone in MusicVAE.

magenta/js/music_rnn/src/model.ts
```diff
@@ -0,0 +1,178 @@
...
+
+    this.lstmCells.length = 0;
+    this.biasShapes.length = 0;
+    let l = 0;
+    while (true) {
+      const cellPrefix = CELL_FORMAT.replace('%d', l.toString());
+      if (!(cellPrefix + 'kernel' in vars)) {
+          break;
+      }
+      this.lstmCells.push(
```
adarob: Can you add a check and a TODO for attention? I think we will silently ignore it now, giving bad results.

magenta/js/music_rnn/src/model.ts
```diff
@@ -0,0 +1,178 @@
...
+    magenta.Sequences.assertIsQuantizedSequence(sequence);
+
+    let continuation: INoteSequence;
+
+    if(!this.initialized) {
+      await this.initialize();
+    }
+
+    dl.tidy(() => {
+      const converterIn = new data.MelodyConverter(
```
adarob: What about drums?

magenta/js/core/src/sequences.ts
cghawthorne: This ended up being a huge pain. Yes, it does turn out that fromObject doesn't do a deep copy unless you first do a toObject, but that's not necessarily available on an INoteSequence.

However, using the 'clone' package on an INoteSequence that is actually a NoteSequence object messes up the prototype somehow. I don't understand exactly what's happening, but the end result is that if your object had a repeated field that was empty before you cloned it, you will not be able to add to that repeated field after you clone it. It has something to do with how the clone package clones the prototype for the object.

Anyway, after trying a bunch of things, I'm now more or less convinced that the easiest/safest way to clone a proto object is to just encode/decode it. Another option would be something like:

```
clone(NoteSequence.toObject(NoteSequence.create(noteSequence)));
```

Or we could add special logic for different behavior depending on whether the object is actually a NoteSequence or just an object that matches INoteSequence.

But I think I prefer the encode/decode option because it works equally well with any kind of INoteSequence and is guaranteed to be a deep copy without any references to the original object. Thoughts?

magenta/js/music_rnn/src/model.ts
cghawthorne: Done

magenta/js/music_rnn/src/model.ts
cghawthorne: Was going to add that in a followup PR.

magenta/js/music_rnn/src/model.ts
adarob: Could you just add something to see if the attention variable is there and throw an error if it is?

PR #1153, opened 2018-05-02
Adding annealed independent blocked Gibbs sampling in the graph for Coconet.

czhuang:


cghawthorne:
You can ignore the py2 test failure, it's fixed in #1155.

PR #1158, opened 2018-05-02
Add documentation and TODO concerning meaning of total_time field in …

adarob:
…NoteSequence.

### DIFF
--- a/magenta/protobuf/music.proto
+++ b/magenta/protobuf/music.proto
@@ -51,9 +51,13 @@ message NoteSequence {
   // A Note combines a MIDI NoteOn and NoteOff into one event with duration.
   repeated Note notes = 8;
   // The total time of the Sequence in seconds.
-  // Total time is defined as the end time of the last note in the sequence.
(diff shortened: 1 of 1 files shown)


magenta/protobuf/music.proto
```diff
@@ -51,9 +51,13 @@ message NoteSequence {
...
+  // Currently the total time is defined as the end time of the last note in the
+  // sequence, and any control changes or rests that occur after the end of the
+  // last note are not included in this time.
+  // In the future, this time will be allowed to extend beyond the last note
+  // end time in order to represent padding. Magenta.js already uses this
+  // interpretation of the field.
+  // TODO(adarob): Update existing code to allow for this new interpretation.
   double total_time = 9;
   // The total time of the sequence in quantized steps.
   // This has the same meaning as the total_time field.
```
cghawthorne: should probably also add a note to total_quantized_steps just to be clear

adarob: Done.

PR #1159, opened 2018-05-03
Coconet: Remove conv weights initializer when not training...

jhowcrof:
(was causing TPU errors and isn't needed for inference) and remove tf.nn.moments from batchnorm when not training (was a bug, shouldn't have been in inference graph).


jhowcrof:
thanks!

PR #1161, opened 2018-05-03
MusicVAE: Add training hparams to configs, add hierarchical encoder configs, and update README.

adarob:


iansimon:
I haven't changed max beta at all but I've used various numbers of free bits.  I'll address it one way or another when I make the checkpoints available.

Issue #1148, opened 2018-04-22
arbitrary_image_stylization_with_weights: command not found


maciej3031:
When I try to run `arbitrary_image_stylization_with_weights` i get: `arbitrary_image_stylization_with_weights: command not found`

The same using: `python arbitrary_image_stylization_with_weights.py` I get:
`
Traceback (most recent call last):
  File "arbitrary_image_stylization_with_weights.py", line 33, in <module>
    from magenta.models.arbitrary_image_stylization import arbitrary_image_stylization_build_model as build_model
ImportError: No module named arbitrary_image_stylization
`

I did everything listed in:
`https://github.com/tensorflow/magenta/blob/master/README.md#manual-install`
and
`https://github.com/tensorflow/magenta/blob/master/magenta/models/arbitrary_image_stylization/README.md`

All other commands work fine. Am I doing something wrong?

akhiljalan:
Same issue here. I think it's an issue with the build, because in the build file we have

```
py_library(
    name = "arbitrary_image_stylization_build_model",
    srcs = [
        "arbitrary_image_stylization_build_model.py",
    ],
    srcs_version = "PY2AND3",
    deps = [
        "//magenta/models/image_stylization:ops",
        ":nza_model",
        ":arbitrary_image_stylization_losses",
        # tensorflow dep
    ],
)
```

Im on Mac OS, and I tried to build in two ways: 

1. Double-clicking the build file, which leads to: 

```
Akhils-MacBook-Air-4:~ akhiljalan$ /Users/akhiljalan/Programming/_wip/magenta-master/magenta/models/arbitrary_image_stylization/BUILD ; exit;
/Users/akhiljalan/Programming/_wip/magenta-master/magenta/models/arbitrary_image_stylization/BUILD: line 19: syntax error near unexpected token `newline'
/Users/akhiljalan/Programming/_wip/magenta-master/magenta/models/arbitrary_image_stylization/BUILD: line 19: `py_binary('
logout
Saving session...
...copying shared history...
...saving history...truncating history files...
...completed.

[Process completed]
```

2. Installing bazel and trying `bazel build BUILD`, which works fine but leads to the same issue as this one 

```
INFO: Analysed target //magenta/models/arbitrary_image_stylization:BUILD (1 packages loaded).
INFO: Found 1 target...
INFO: Elapsed time: 0.307s, Critical Path: 0.02s
INFO: Build completed successfully, 1 total action
```

CodeW1zard:
[...]

PR #1189, opened 2018-05-31
Fix for issue 1148

bjaress:
* Includes arbitrary stylization scripts in pip build

I wanted to use these scripts and ran into the issue here: <https://github.com/tensorflow/magenta/issues/1148>.  This seems to fix it for a local build.


bjaress:
I signed it!

cghawthorne:
Thanks!

PR #1221, opened 2018-06-30
Coconet: split pointwise to reduce size

czhuang:


cghawthorne:
For py3 compatibility, I think you'll need to switch to six.iteritems

PR #1230, opened 2018-07-13
add link to performance_with_dynamics_and_modulo_encoding bundle

iansimon:
@vidavakil Sorry it took me so long to get to this!


vidavakil:
Thank you very much Ian, and Adam, for including the bundle with Modulo-Encoding.

PR #1239, opened 2018-07-27
Next frame prediction based on pix2pix

dh7:
This PR add some code to generate video using pix2pix recursively in a folder named "video".
It dynamically import pix2pix using Bazel.

It is essentially a script to extract frame from a video and train pix2pix out of it, then generate some video a several stage of the training.


dh7:
i just accepted the CLA for my other email address

PR #1243, opened 2018-07-30
Cast rate to float when computing sampling probability.

adarob:


adarob:
@cghawthorne also bumping to 0.3.10

PR #1250, opened 2018-08-10
Random note sequence augmentation

daphnei:
Common library calls for transposing sequence by specified amount and for randomly augmenting a sequence.


daphnei:
I signed it!

daphnei:
I signed it!

PR #1257, opened 2018-08-14
Update docker for v0.3.11 release

### LINKED ISSUE #1242: Docker container does not contain Music VAE model
It would be nice if you  could update it  due to I have some troubles with environment set up

cghawthorne:
Fixes #1242.

### DIFF
--- a/magenta/tools/docker/Dockerfile
+++ b/magenta/tools/docker/Dockerfile
@@ -1,10 +1,10 @@
-FROM tensorflow/tensorflow:1.4.1-devel
+FROM tensorflow/tensorflow:latest-devel
 
 MAINTAINER Curtis Hawthorne <fjord@google.com>
 
 # Download and build Magenta.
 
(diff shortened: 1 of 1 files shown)


cghawthorne:
@AndreyShagal Can you provide a little more context? Is there a longer error message or a stack trace? Does this happen for training or inference?

AndreyShagal:
I updated docker package.
And after that any execution (training/) gives only one line "Illegal instruction"   no stack trace or any other error messages. For example - music_vae_train (or any other method)  execution with or without parameters gives "Illegal instruction"

cghawthorne:
This may be related to https://stackoverflow.com/questions/49094597/illegal-instruction-core-dumped-after-running-import-tensorflow. Do you have an older CPU that doesn't support AVX instructions? You may have to either manually build tensorflow or pip install an older version of tensorflow.

PR #1270, opened 2018-08-24
Check-in Latent Transfer Model

alantian:
Check-in of Latent Transfer model and training scripts.

This check-in contains the core parts of Latent Transfer model including:

- The transfer model using Variational Inference that bridges latents spaces of data space models.
- The scripts for training such model.
- Related configs which also allow parameterization from command line.
- Training of VAE and classifier of data space (MNIST and CelebA)
- The utility for associated data space model on more data (Fashion MNIST, WaveGAN) and on more functionality (sampling, encoding).


alantian:
I signed it!

PR #1288, opened 2018-09-13
Implements TranspositionToPipeline

veskoch:
TranspositionToPipeline is useful for data standardisation and is modelled after TranspositionPipeline which has been used for data augmentation.


veskoch:
Just signed the CLA!

PR #1311, opened 2018-10-17
Update uses of tf.contrib.distributions to use TensorFlow Probability.

srvasude:
Given that tf.contrib.distributions is being deprecated/removed, update all uses of tf.contrib.distributions to use tfp.distributions (TensorFlow Probability: https://www.tensorflow.org/probability/).


adarob:
Please also add tensorflow_probability to the required pip packages in https://github.com/tensorflow/magenta/blob/master/magenta/tools/pip/setup.py. Thanks!

srvasude:
Added TFP (GPU/non-GPU) installation to setup.

PR #1284, opened 2018-08-31
image_stylization_create_dataset.py 

lisiping0817:
this file maybe lack of console_entry_point method ,so when we from magenta.models.image_stylization.image_stylization_create_dataset import console_entry_point ,there will a 
 importerror.

Issue #1279, opened 2018-08-30
magenta/magenta/models/image_stylization/


lisiping0817:
![image](https://user-images.githubusercontent.com/41105367/44827987-6a8c4300-ac48-11e8-92dc-f8878f1ee00f.png)
  

i think this code maybe have mistake   so we cannot  import name console_entry_point.
so  i want to know the truth   thank you

lisiping0817:
![image](https://user-images.githubusercontent.com/41105367/44828095-dbcbf600-ac48-11e8-9e64-fd7873ca3b62.png)

lisiping0817:
code in the image stylization create dataset

cghawthorne:
Yes, it looks like it should have a `console_entry_point` method similar to the other commands in that directory. Can you try adding that and send a pull request?

lisiping0817:
thank you

PR #1320, opened 2018-10-29
fix: added missing console_entry_point definition

ema987:
Fixes error described in #1279

Also refer to [PR1284](https://github.com/tensorflow/magenta/pull/1284) which could be closed.


ema987:
I signed it!

ema987:
Thanks to you!

PR #1324, opened 2018-10-30
correct for 1-based indexing of MusicNet program numbers

iansimon:


magenta/music/musicnet_io.py
```diff
@@ -54,7 +54,7 @@ def note_interval_tree_to_sequence_proto(note_interval_tree, sample_rate):
     note.velocity = MUSICNET_NOTE_VELOCITY
     note.start_time = float(note_interval.begin) / sample_rate
     note.end_time = float(note_interval.end) / sample_rate
-    note.program = note_data[0]
+    note.program = note_data[0] - 1
```
adarob: Can you add a comment somewhere explaining why you're doing this?

PR #1329, opened 2018-10-31
add piano genie visibility

iansimon:


magenta/models/piano_genie/BUILD
```diff
@@ -66,6 +66,9 @@ py_binary(
     srcs = [
         "eval.py",
     ],
+		visibility = [
```
cghawthorne: nit: tabbing is inconsistent with the rest of the file

iansimon: Oops, not sure how that happened.  Does it look right now?

PR #1338, opened 2018-11-05
python 3 compatibility for music vae

TheAustinator:
Converted an instance of range() to list(range()) because insert was subsequently called on the output, which is a list in python 2, but a generator in python 3, and thus an error is raised. By converting it to a list, the error will not be raised in either version. 

Converted an instance of dict.iteritems() to dict.items() for compatibility with both python 2 and 3. There are other instances of dict.iteritems() that did not cause problems in my code that may cause problems for others, but since iteritems() returns an iterator and items() returns a list, testing will be required before changing these instances.


adarob:
Thanks! Looks good once you sign the CLA.

TheAustinator:
Great. I signed it

PR #1368, opened 2018-11-30
Fixes for onsets and frames testing

faraazn:
Hi again! I was running the code for onsets and frames testing and ran into a couple bugs.

First, the call to `evaluate_once` was missing the `master` argument (see this [link](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/contrib/slim/python/slim/evaluation.py)). It's a little confusing because the comments for `evaluate_once` in this file differ from the actual implementation.

Second, when printing out the metrics I found that not all of the metrics were scalars, and this caused errors while printing them out. I added a check to resolve this.

Lastly, just wanted to mention that are a few more python 3 compatibility issues. They mainly have to do with the fact that calls to values() and keys() on `dict`s don't return lists (ex. lines 181, 182, 184). Adding `list` wrappers to these calls resolved the compatibility issues for me.

### DIFF
--- a/magenta/models/onsets_frames_transcription/train_util.py
+++ b/magenta/models/onsets_frames_transcription/train_util.py
@@ -152,7 +152,7 @@ def evaluate(train_dir,
         tf.contrib.training.SummaryAtEndHook(eval_dir)]
     tf.contrib.training.evaluate_repeatedly(
         train_dir,
-        eval_ops=metrics_to_updates.values(),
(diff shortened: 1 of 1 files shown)


adarob:
Could you also make the py3 changes here or in a separate PR?

faraazn:
Yeah, added them in another commit here. Wasn't sure if the change would be backwards compatible but looks like it is.

adarob:
Thanks again!

PR #1380, opened 2018-12-13
Open sourcing model distillation

### LINKED ISSUE #1373: arbitrary_image_stylization code for distillation
Would you be interested in adding the scripts I used for distillation (stylenet inceptionv3->mobilenet) + another thing i'm working on where i replace the transformnet conv2d layers with separable_conv2d?

Right now i haven't uploaded them publicly since they're in pretty messy colab notebooks but I could clean it up and have something like `distill_arbitrary_image_stylization_to_mobilenet.py`. This way, other people with more resources than me could run them (I can only do around 80K training steps in a few days with Google Colab) and probably get better models. If not, then I'll probably just dump the notebooks as is in my repo.

reiinakano:
~WIP as I haven't tested this yet. I use the COCO dataset (content) + PBN dataset + Describable texture dataset in Colab using the `tf.Dataset` API as there's no easy way to get ImageNet in Colab AFAIK. For now, putting this up for some early comments.~

edit: Not WIP anymore pending the Mobilenet import. Had to change some `iteritems` to `items` for Python3

Fixes #1373


magenta/models/arbitrary_image_stylization/arbitrary_image_stylization_build_mobilenet_model.py
```diff
@@ -0,0 +1,149 @@
...
+import tensorflow as tf
+
+from magenta.models.arbitrary_image_stylization import arbitrary_image_stylization_losses as losses
+from magenta.models.arbitrary_image_stylization import nza_model as transformer_model
+from magenta.models.image_stylization import ops
+
+try:
+  from nets.mobilenet import mobilenet_v2, mobilenet
+except ImportError:
+  print("Cannot import MobileNet model. Make sure to install slim models library described "
```
reiinakano: What's a good way for fixing this?

reiinakano:
Not WIP anymore pending the Mobilenet import. Seems to work. Had to change some `iteritems` to `items` for Python3

adarob:
@vdumoulin can you review this since it touches your code?

magenta/models/arbitrary_image_stylization/arbitrary_image_stylization_build_mobilenet_model.py
adarob: does it exist as a pip package?

reiinakano: I don't think it does. The installation instructions are here and make no mention of a pip package. https://github.com/tensorflow/models/tree/master/research/slim

magenta/models/arbitrary_image_stylization/README.md
```diff
@@ -252,3 +252,24 @@ $ CUDA_VISIBLE_DEVICES= arbitrary_image_stylization_evaluate \
       --logtostderr
 ```
 
+## Distill style prediction network using MobileNetV2
+
+To distill the InceptionV3 style prediction network using a MobileNetV2 model,
+you will need the MobilenetV2 pre-trained checkpoint from 
+https://storage.googleapis.com/mobilenet_v2/checkpoints/mobilenet_v2_1.0_224.tgz.
+You will also need the checkpoint from a trained arbitrary image stylization model
+(You can use the one available here).
```
vdumoulin: How about adding a link to the checkpoint?

magenta/models/arbitrary_image_stylization/arbitrary_image_stylization_build_mobilenet_model.py
```diff
@@ -0,0 +1,154 @@
...
+                          mobilenet_trainable=True,
+                          style_params_trainable=False,
+                          transformer_trainable=False,
+                          reuse=None,
+                          mobilenet_end_point='layer_19',
+                          style_prediction_bottleneck=100,
+                          adds_losses=True,
+                          content_weights=None,
+                          style_weights=None,
+                          total_variation_weight=None):
```
vdumoulin: For functions defined in this module, can you add docstrings explaining what each argument is for? For instance, it's not immediately clear what behavior to expect if `style_prediction_bottleneck <= 0`, which judging from line 118 is supported.

magenta/models/arbitrary_image_stylization/arbitrary_image_stylization_build_model.py
```diff
@@ -130,6 +131,8 @@ def style_prediction(style_input_,
     style_prediction_bottleneck: int. Specifies the bottleneck size in the
         number of parameters of the style embedding.
     reuse: bool. Whether to reuse model parameters. Defaults to False.
+    inception_only: bool. If true, does not build the mapping from bottleneck
```
vdumoulin: Is there a reason why one would need to avoid constructing that part of the computation graph rather than ignoring the `style_params` dict returned by this function?

reiinakano:
Done. For the docstrings, I just copied over the relevant description from the original functions I based this on.

magenta/models/arbitrary_image_stylization/arbitrary_image_stylization_build_model.py
reiinakano: No reason other than I didn't think of it :)

adarob:
@reiinakano please add the binary to https://github.com/tensorflow/magenta/blob/52e6ca806df312ae10d4e32852ecae9675cc4996/setup.py#L67 and this will be good to go

reiinakano:
What about the slim-models dependency? Right now the only way people can use it is to clone it and ensure the module is in the path.

adarob:
> What about the slim-models dependency? Right now the only way people can use it is to clone it and ensure the module is in the path.

Ah, please explain this in the README. I don't think there is a better solution unfortunately, aside from writing a script to do this for the user.

PR #1391, opened 2018-12-22
Optimize onsets and frames create_dataset

faraazn:
Hey there!

I noticed that the onsets and frames data processing was taking a lot longer than I expected. With some rough profiling I found that most of the runtime was being taken up by calls to `audio_io.wav_data_to_samples`. Additionally, I found that calls to `audio_io.crop_wav_data` were resulting in redundant calls to `wav_data_to_samples`. So I came up with this fix.

I ran a few tests directly comparing the old and new values of `new_wav_data` to verify that this code is functionally equivalent to the original. Running `generate_train_set` with `TRAIN_DIRS=['AkPnBcht/MUS']`, I got the following results with my gtx1080ti intel i7 desktop at home:

- Original code: **57 min 22.20 sec**
- Optimized code: **2 min 56.36 sec**

30 files, processed into 371 chunks.

This is around a 20x speedup :) On another dataset, this brought my overall processing time (including test data) down from 8 hrs to 45 min.

Also included in this commit are a couple more py3 compatibility fixes, and one aesthetic change that I found useful!

### DIFF
--- a/magenta/models/onsets_frames_transcription/onsets_frames_transcription_create_dataset.py
+++ b/magenta/models/onsets_frames_transcription/onsets_frames_transcription_create_dataset.py
@@ -75,18 +75,19 @@ def generate_train_set(exclude_ids):
                                    'maps_config2_train.tfrecord')
 
   with tf.python_io.TFRecordWriter(train_output_name) as writer:
-    for pair in train_file_pairs:
-      print(pair)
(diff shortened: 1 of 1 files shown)


adarob:
Awesome, thanks! Can you please confirm that it also works for py2?

faraazn:
Yup, ran it on a py2 environment and it works!

PR #1403, opened 2019-01-08
Full lint for google3 re-import.

adarob:


magenta/models/latent_transfer/interpolate_joint.py
```diff
@@ -23,15 +23,15 @@
...
-import tensorflow as tf
+import os
 
 from magenta.models.latent_transfer import common
 from magenta.models.latent_transfer import common_joint
 from magenta.models.latent_transfer import model_joint
+import numpy as np
+import tensorflow as tf
 
+join = os.path.join
```
iansimon: There are few enough calls that I would prefer to just use the full `os.path.join` for each call.

magenta/models/latent_transfer/sample_wavegan.py
```diff
@@ -21,14 +21,15 @@
...
-from os.path import join
+import operator
+import os
 
 import numpy as np
 from scipy.io import wavfile
 import tensorflow as tf
 from tqdm import tqdm
 
+join = os.path.join
```
iansimon: Same as above.

magenta/models/latent_transfer/train_joint.py
```diff
@@ -27,13 +27,12 @@
 import importlib
 from os.path import join
```
iansimon: Why is from `os.path import join` used here but not in the previous files?  I prefer the full `os.path.join` either way.

magenta/protobuf/music_pb2.py
```diff
@@ -3,11 +3,13 @@
...
 
 import sys
-_b=sys.version_info[0]<3 and (lambda x:x) or (lambda x:x.encode('latin1'))
+
 from google.protobuf import descriptor as _descriptor
 from google.protobuf import message as _message
 from google.protobuf import reflection as _reflection
 from google.protobuf import symbol_database as _symbol_database
+
+_b=sys.version_info[0]<3 and (lambda x:x) or (lambda x:x.encode('latin1'))
```
iansimon: How sure are you that this doesn't break anything?  Is there any chance the above imports were somehow reading this variable?

magenta/models/latent_transfer/train_joint.py
adarob: Weird. I'm not sure how the linter missed this...

magenta/protobuf/music_pb2.py
adarob: reverted

Issue #1420, opened 2019-01-29
setup.py test failing with syntaxerror


vitorarrais:
Error:
```
Traceback (most recent call last):
  File "setup.py", line 154, in <module>
    tests_require=['pytest', 'pylint'],

  (...)

  File "/Users/vitorarrais/magenta/.eggs/pytest_pylint-0.14.0-py2.7.egg/pytest_pylint.py", line 13, in <module>
    from pylint import lint
  File "/Users/vitorarrais/magenta/.eggs/pylint-2.3.0.dev1-py2.7.egg/pylint/lint.py", line 218
    def _cpu_count() -> int:
                     ^
SyntaxError: invalid syntax
```

Occurred when running `python setup.py test` in Python 2.7 environment

Any thoughts?

cghawthorne:
@adarob I think you mentioned that pytest only works in python 3. Should we modify setup.py to behave differently for py2 or something?

PR #1421, opened 2019-01-30
require pylint < 2.0 for py2 envs

vitorarrais:
## Description
pylint >= 2.x breaks when `setup.py test` runs against a py2 environment

## Solution
force `setup.py` to use pylint < 2.0 for py2 envs 

Closes issue [#1420](https://github.com/tensorflow/magenta/issues/1420)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -49,6 +49,7 @@ REQUIRED_PACKAGES = [
     'pretty_midi >= 0.2.6',
     'protobuf',
     'pygtrie >= 2.3',
+    'pylint < 2.0.0;python_version=="2.7"',
     'python-rtmidi >= 1.1, < 1.2',  # 1.2 breaks us
     'scipy >= 0.18.1',
     'sk-video',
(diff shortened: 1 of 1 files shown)


cghawthorne:
Actually, @vitorarrais, is there a way to have the pylint py2 requirement specified only in the `test_requires` portion of setup.py? We don't necessarily want to require pylint for all python 2 installations of the package.

vitorarrais:
@cghawthorne done.

setup.py
```diff
@@ -150,5 +150,5 @@
         'magenta': ['models/image_stylization/evaluation_images/*.jpg'],
     },
     setup_requires=['pytest-runner', 'pytest-pylint'],
-    tests_require=['pytest', 'pylint'],
+    tests_require=['pytest', 'pylint < 2.0.0;python_version=="2.7"'],
```
cghawthorne: Does this maintain the requirement for py3? I'm not super familiar with how [environment markers](https://www.python.org/dev/peps/pep-0508/#environment-markers) work, but should it be something like:

```
'pylint < 2.0.0;python_version<"3"',
'pylint > 2.0.0;python_version>="3"',
```

vitorarrais: I'm not familiar with this syntax either. I think you're right, even though the missing part for py3 doesn't break tests for py3 environment. This is odd. Anyway, thank you for pointing out!

PR #1488, opened 2019-03-30
Open audio file in binary mode

longern:
Fix UnicodeDecodeError in onsets_frames_transcription_transcribe.py


longern:
I signed it!

Issue #1477, opened 2019-03-16
Automated install script contains python=27 spec


nicchan:
I followed the Automated Install (w/ Anaconda) in the README.md.
I am using python 3.6.x.
However, under the package plan, I saw that the `python=2.7` spec was added during the operation of the installation script.

I believe it's coming from line 88 of the [***magenta-install.sh***](https://github.com/tensorflow/magenta/blob/master/magenta/tools/magenta-install.sh):
[conda create -n magenta python=2.7](https://github.com/tensorflow/magenta/blob/2cbc96922b7933549983a8d8845f46002974becb/magenta/tools/magenta-install.sh#L88)

Is this okay? or do I need to change anything for python 3.6.x compatibility?
(I also don't want to use python 27 because its no longer supported in 2020)

Thanks!!

log below from my console:  
```
=========================================
anaconda detected, skipping conda install
=========================================


==============================
setting up magenta environment
==============================

Solving environment: done


==> WARNING: A newer version of conda exists. <==
  current version: 4.4.10
  latest version: 4.6.8

Please update conda by running

    $ conda update -n base conda



## Package Plan ##

  environment location: /Users/Nicholas/anaconda3/envs/magenta

  added / updated specs: 
    - python=2.7
```

cghawthorne:
Magenta should be compatible with Python 3. Would you want to try updating the auto install script to use py3 and submit a PR if that works?

PR #1500, opened 2019-04-11
Update magenta-install.sh

falaktheoptimist:
As per #1477 anaconda default changed to python3

### DIFF
--- a/magenta/tools/magenta-install.sh
+++ b/magenta/tools/magenta-install.sh
@@ -61,7 +61,7 @@ if [[ ! $(which conda) ]]; then
     echo "==========================================="
     echo ""
     readonly CONDA_INSTALL="/tmp/${MINICONDA_SCRIPT}"
-    readonly CONDA_PREFIX="${HOME}/miniconda2"
+    readonly CONDA_PREFIX="${HOME}/miniconda3"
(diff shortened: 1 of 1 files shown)


magenta/tools/magenta-install.sh
```diff
@@ -85,7 +85,7 @@ echo "setting up magenta environment"
 echo "=============================="
 echo ""
 
-conda create -n magenta python=2.7
+conda create -n magenta python=3.6
```
cghawthorne: should probably also change the download above to get Miniconda3 instead of Miniconda2.

falaktheoptimist: Done. Also, pip changed to pip3.

cghawthorne: one other nit: can you change the path for CONDA_PREFIX to also use miniconda3?

falaktheoptimist: Done

cghawthorne: Let's actually move this to python=3.7 since that's what we do our Travis tests with

falaktheoptimist: Done.

PR #1512, opened 2019-04-18
Lock tensorflow to working 1.13.1 version

sbrother:
When following the "Development Environment" instructions in a clean environment (tested on Ubuntu 18.04, MacOS, py3 and py2), we were consistently getting 62 failures, mostly with the traceback `AttributeError: module 'tensorflow' has no attribute 'contrib'`. It looks like this module was removed in Tensorflow 2.0, which is currently pulled in by `python setup.py install`. Locking the `tensorflow` and/or `tensorflow-gpu` versions to 1.13.1 fixes the issue.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -62,9 +62,9 @@ REQUIRED_PACKAGES = [
 ]
 
 if gpu_mode:
-  REQUIRED_PACKAGES.append('tensorflow-gpu >= 1.12.0')
+  REQUIRED_PACKAGES.append('tensorflow-gpu == 1.13.1')
 else:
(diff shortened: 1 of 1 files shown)


cghawthorne:
This shouldn't be necessary. I think the thing we need to change is to switch our dev environment setup instructions to use `pip install -e .` instead of setup.py. pip handles dependencies better than setup.py, including not pulling in prerelease versions. I'll send a PR for that shortly.

cghawthorne:
Closing this PR in favor of the updated instructions in 5c60974fa800429f55839038f6b5d51f905fb122

Issue #1360, opened 2018-11-25
Organize magenta.music into library.


adarob:
We should start by re-organizing the functions in sequences_lib, etc. and exporting them directly under magenta.music.

cifkao:
+1. There are functions and data structures in Magenta that could be very useful to the MIR community, but it can be impractical to have Magenta as a dependency. For example, I would like to use the `NoteSequence` proto to store some pre-processed MIDI files, but I can't install Magenta because I need to use an older version of TensorFlow.

If there's any way I can help with this, I can look into it.

adarob:
Thanks for the offer! Like I mention here, a first step would be refactoring what's under magenta/music and exposing it (for now) as a library within magenta. That would make it easier to pull out later and put into its own package. Of course, totally separating it from tensorflow dependencies might be a challenge still...

cghawthorne:
Most of the tensorflow dependencies are around tf.logging, which we can easily replace with abseil, and tf.gfile, which is more of a problem since we need it for internal compatibility. I suppose we could do something like write our own magenta.gfile wrapper, but that seems a little hacky.

cifkao:
Maybe I misunderstood what "library" meant in this context (presumably it's about creating a Bazel rule and not a separate Python package). If that's the case, sorry for spamming this issue and I should probably create a new one.

adarob:
No, it does mean a separate package. But before we can do that we first need to break all of the internal dependencies.

cifkao:
OK, thanks for clarifying. So the goal would be to move the code (everything under `music` and `protobuf`?) to a new directory in this repo (eventually with its own `setup.py`), is that correct?

I tried to find things that `music` depends on. The only ones which I don't really see how to resolve are `pipeline.statistics` and `common.testing_lib`, which are imported in `music.*_lib` or the associated tests. On the other hand, many modules (notably `music.*_io` plus the protobuffers) seem to be already self-contained. I'm wondering if it would be reasonable to start by moving out the self-contained or easy-to-resolve things and leave the rest for later.

cifkao:
I gave it a stab (see #1519). The only dependency left seems to be `pipeline.statistics`; I'm not sure what it does, but it doesn't seem to be music-specific enough to move it under `music`.

adarob:
Thanks! @cghawthorne can you help review this?
Doing this will also allow us to put MAESTRO in TFDS.

adarob:
[...]

PR #1519, opened 2019-04-28
Towards an independent magenta.music

cifkao:
Untangling some dependencies to clear the way for factoring out `magenta.music` as a separate package (#1360):

* Moving `music.encoder_decoder.EncoderPipeline` to `pipelines.event_sequence_pipeline` (maybe it could be renamed to `GenericEncoderPipeline` or something and moved to `pipelines_common` instead)
* Moving `make_sequence_example` from `common.sequence_example_lib` to `music.encoder_decoder`
* Copying `parse_test_proto` from `common.testing_lib` to `music.testing_lib`

### DIFF
--- a/magenta/common/__init__.py
+++ b/magenta/common/__init__.py
@@ -21,5 +21,4 @@ from .nade import Nade
 from .sequence_example_lib import count_records
 from .sequence_example_lib import flatten_maybe_padded_sequences
 from .sequence_example_lib import get_padded_batch
-from .sequence_example_lib import make_sequence_example
 from .tf_utils import merge_hparams
--- a/magenta/common/sequence_example_lib.py
+++ b/magenta/common/sequence_example_lib.py
@@ -15,39 +15,11 @@
 """Utility functions for working with tf.train.SequenceExamples."""
 
 import math
-import numbers
 
 import tensorflow as tf
 
 QUEUE_CAPACITY = 500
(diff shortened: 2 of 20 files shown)


cifkao:
There is still the dependency on `pipelines.statistics`, not sure what to do about that. Also should I move `protobuf` inside `music`?

As for the CLA, hopefully I'll be able to have it signed next week.

cifkao:
CLA signed!

cghawthorne:
Overall, I think this looks good. One thought I had is that I'm wondering whether or not the encoder/decoder stuff should eventually end up in our separate magenta.music library. It's pretty specific to our *_rnn models, and I think it might make more sense to just move those files to models/shared.

@iansimon what do you think?

iansimon:
Agree that the encoder_decoder-named files should not go into magenta music, at least for now.  Later on if some of it is really fundamental it can be moved back.

cifkao:
Are you thinking that these should be moved out of `music` as part of this pull request?

iansimon:
I think it's fine to leave them where they are for now.  @cghawthorne what do you think?

cghawthorne:
Yeah, that can wait for another PR.

cghawthorne:
I'm not sure why the Travis CI tests aren't running. Can you maybe try rebasing your PR against HEAD and see if that triggers it?

Issue #1527, opened 2019-05-11
magenta_midi.py does not work accordingly, after following setup & launching instructions


SamMichalik:
Hello,
I am trying to get magenta_midi.py running on Ubuntu 18.04, but have been unable to do so,
after following the instructions in the README file.

After following every step and launching magenta_midi.py with the `input_ports`, `output_ports` and `bundle_files` parameters specified, the program runs, but does not produce any sound.

I was able to achieve half-success, after modifications in the qjackctl -> Connections -> ALSA window, where I made the following links

- VMPK Output ---> Fluid Synth (which made audible my inputs on the controller)
- VMPK Output ---> RtMidiIn Client (without this, the model apparently didn't receive input, because no responses were produced while the metronome was audible)
- RtMidiOut Client ---> Fluid Synth (this made the metronome and the model's responses audible) 

Unfortunately, after the first response from the model, the RtMidiIn Client, disappears from the list of input ports in the aforementioned ALSA window and no additional responses from the model are produced.

SamMichalik:
I think I may have solved it. In MidiHub's constructor the `_timestamp_and_handle_message` callback is attached to `inport`, however `inport` is only a local variable and is never assigned anywhere, so the port instance gets garbage collected later on, while the interaction is still active. 

Starting on line 904 in `midi_hub.py` :
```
    if input_midi_ports:
      for port in input_midi_ports:
        if isinstance(port, mido.ports.BaseInput):
          inport = port
        else:
          virtual = port not in get_available_input_ports()
          if virtual:
            tf.logging.info(
                "Opening '%s' as a virtual MIDI port for input.", port)
          inport = mido.open_input(port, virtual=virtual)
        # Start processing incoming messages.
        inport.callback = self._timestamp_and_handle_message
```
Additionally, `input_midi_ports` elements are strings, so the `isinstance` test always fails and all ports are in effect virtual.

If you want, I can open a pull request to fix this.

adarob:
Please do open a PR. Thanks!

*From: *Samuel Michalik <notifications@github.com>
*Date: *Mon, May 13, 2019, 6:38 AM
*To: *tensorflow/magenta
*Cc: *Subscribed

I think I may have solved it. In MidiHub's constructor the

PR #1528, opened 2019-05-13
Fix issue #1527

SamMichalik:
- Open input ports in `MidiHub` are no longer prematurely deleted.
- Determination of port existence has been corrected, so only names referring to non-existing ones result in the creation of new virtual ports.

### DIFF
--- a/magenta/interfaces/midi/midi_hub.py
+++ b/magenta/interfaces/midi/midi_hub.py
@@ -901,32 +901,30 @@ class MidiHub(object):
 
     # Open MIDI ports.
 
+    inports = []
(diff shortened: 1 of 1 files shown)


SamMichalik:
I signed it!

adarob:
It looks like this is failing on Travis: https://travis-ci.org/tensorflow/magenta/builds/531936954

 @SamMichalik can you take a look?

SamMichalik:
Sorry about that. Only the first point of my issue was valid; I wrongly assumed by reading the sources, that only strings are ever passed as port parameters, which wasn't the case in the tests.

PR #1531, opened 2019-05-15
Fixes + documentation to run Coconet code with Python 3

everettk:
These changes were required for me to run the `sample_bazel.sh` and `train_bazel.sh` scripts in the Coconet project, using Python 3, and I tried to improve the README to make it clearer to someone new to the project.

I haven't contributed to this project before, but I thought it would be helpful to submit a PR! Happy to incorporate any comments/feedback.

**Testing notes:**
- Follow the usage instructions in the README in this PR.
- If you have a pip-installed version of magenta, set your $PYTHONPATH to the location of the magenta directory from this branch, so that the local version of magenta (containing these changes) takes precedence when magenta is imported.
- The pre-trained model downloaded from [here](http://download.magenta.tensorflow.org/models/coconet/checkpoint.zip) requires a fix in order to work. The download contains a checkpoint file named `checkpoint`, which needs to point to the checkpoint binaries in that directory. Change the `checkpoint` file to contain only the line `model_checkpoint_path: "best_model.ckpt"`.

**To Do:**
- [ ] This doesn't block this PR because it only uses the `.npz` file, but the pickled files in the [repo](https://github.com/czhuang/JSB-Chorales-dataset) containing the full JSB chorale datasets, can't be unpickled using Python 3 without some trickery, because apparently Python 3 doesn't like to unpickle files that Python 2 pickled. Would be helpful to submit a separate PR containing a Python 3-compatible version of those files.


everettk:
I signed it!

magenta/models/coconet/lib_data.py
```diff
@@ -172,7 +172,7 @@ def __init__(self, **kwargs):
     """
     assert set(kwargs.keys()) == self.keys
     assert all(
-        len(value) == len(kwargs.values()[0]) for value in kwargs.values())
+        len(value) == len(list(kwargs.values())[0]) for value in list(kwargs.values()))
```
czhuang: To fix a line too long lint error:
```suggestion
        len(value) == len(list(kwargs.values())[0])
        for value in list(kwargs.values()))
```

sleep-yearning:
Thanks for the work so far! This has been a problem for us too currently.
Is there anything left to do or could this be pulled now with the requested change?

magenta/models/coconet/lib_data.py
sleep-yearning: I implemented this and made a PR in @everettk 's repo. 
However, since it's not pulled yet, you could pull it from my repo instead, since I can't suggest an edit on this PR directly. 
Please advise if I should make a separate PR or how to proceed. :)

PR #1547, opened 2019-06-18
fix bug in linear_mel_weight_matrix

Caoang327:
According to the theory, linear segments are linear in the mel domain, not Herze. So the slopes should be calculated in the mel domain. The code of tensorflow/mel_ops and librosa/filters/mel calculate it on mel domain as well.
Thanks


Caoang327:
I signed it!

Issue #1558, opened 2019-07-04
Linter throwing CS errors


ivanproskuryakov:
`pytest` command throwing a set of errors related to CS, might be worth to fix tho.

### pytest output
```
___________________________________________________________________________ [pylint] magenta/models/coconet/coconet_evaluate.py ____________________________________________________________________________
R1720: 73 main: Unnecessary "else" after "raise" [no-else-raise]
________________________________________________________________________________ [pylint] magenta/music/performance_lib.py _________________________________________________________________________________
R1720:839 NotePerformance._from_quantized_sequence: Unnecessary "else" after "raise" [no-else-raise]
________________________________________________________________________________ [pylint] magenta/pipelines/dag_pipeline.py ________________________________________________________________________________
R1720:301 DAGPipeline.__init__: Unnecessary "else" after "raise" [no-else-raise]
R1720:330 DAGPipeline.__init__: Unnecessary "else" after "raise" [no-else-raise]
____________________________________________________________________ [pylint] magenta/video/next_frame_prediction_pix2pix/join_pairs.py ____________________________________________________________________
C0411: 25 : standard import "import os" should be placed before "import ntpath" [wrong-import-order]
C0411: 26 : standard import "from random import shuffle" should be placed before "import ntpath" [wrong-import-order]
_______________________________________________________________________________ [pylint] magenta/video/tools/convert2jpg.py ________________________________________________________________________________
C0411: 21 : standard import "import os" should be placed before "import ntpath" [wrong-import-order]
_______________________________________________________________________________ [pylint] magenta/video/tools/random_pick.py ________________________________________________________________________________
C0411: 24 : standard import "import os" should be placed before "import ntpath" [wrong-import-order]
C0411: 25 : standard import "import random" should be placed before "import ntpath" [wrong-import-order]
C0411: 26 : standard import "import shutil" should be placed before "import ntpath" [wrong-import-order]
=====================================================
failed, 738 passed, 318 skipped, 47 warnings in 170.86 seconds =====================================================
```

<img width="997" alt="Screen Shot 2019-07-09 at 2 18 17 PM" src="https://user-images.githubusercontent.com/3517364/60883989-6b243b80-a254-11e9-96af-cce95113e2a3.png">

### env
```
os: macOs Mojave 10.14 (18A391)
conda: 4.6.11
platform darwin -- Python 3.7.3, pytest-4.3.1, py-1.8.0, pluggy-0.9.0 -- /anaconda3/bin/python
[...]

PR #1561, opened 2019-07-09
Linter throwing CS errors #1558

ivanproskuryakov:
https://github.com/tensorflow/magenta/issues/1558

### before
```
6 failed, 738 passed, 318 skipped, 47 warnings in 170.86 seconds
```

### after

```
741 passed, 321 skipped, 47 warnings in 181.72 seconds
```


ivanproskuryakov:
@adarob Probably there's a mismatch between the version python on CI and my local mac, which makes test to fail, python 3.5 vs python 3.7 and its safer to close the PR, and do not merge it at all.
Also, master is broken.

PR #1570, opened 2019-07-17
Finish separating magenta.music from magenta

### LINKED ISSUE #1360: Organize magenta.music into library.

cifkao:
See #1360.

Important changes:
- Moving ~~all protobuffers~~ the `music` protobuffer to `magenta.music.protobuf`.
- ~~Making `magenta` a [namespace package](https://packaging.python.org/guides/packaging-namespace-packages/), which means removing all the imports from the top-level `__init__.py`. I don't know what the consequences are, but the tests are passing.~~
- ~~`magenta.music` now lives in `magenta-music` and has its own `setup.py`, the rest of the code stays in the root of the repo.~~


cifkao:
The linter error seems unrelated...

cifkao:
Rebased to master, now the tests are passing.

I tied the versions of the two libraries together, not sure if it's a good idea.

adarob:
Sorry for the delay...looking at this today.

magenta-music/.pylintrc
```diff
@@ -0,0 +1,219 @@
+[MASTER]
```
adarob: I think you can leave this file out and it will default to the other one.

cifkao: I guess, unless pytest copies the package to some temporary location.

Let me know if any action is required from me.

cifkao:
@adarob I moved `magenta.music` back in the main package as requested but kept the other changes (moving protobuffers and test files). I think this should be ready to merge now.

You might want to re-generate `__init__.py` (I edited it by hand).

adarob:
trying to import again ... stay tuned!

PR #1580, opened 2019-08-22
Added mobile version to style transfer models

khanhlvg:
* Added mobile version of arbitrary_image_stylization model and a script to convert it to TF Lite
* Re-added the mobile image_stylization model that was deleted by this [commit](https://github.com/tensorflow/magenta/commit/610718c4e5072a98ccb79fbca4b58febb5c6267f)


ClarkChin08:
@khanhlvg hi i noticed tf_hub has add a V2 style transfer model, but the pretrained checkpoint seems still the V1 model, can you share the V2 pretrained checkpoint?

PR #1590, opened 2019-09-21
Changing deprecated instruction

bsmnyk:
tf.python_io.TFRecordWriter  to tf.io.TFRecordWriter


bsmnyk:
@googlebot I signed it!

PR #1591, opened 2019-09-21
Updating deprecated Instruction

bsmnyk:
Changed tf.logging.info to tf.compat.v1.logging.info

### DIFF
--- a/magenta/pipelines/statistics.py
+++ b/magenta/pipelines/statistics.py
@@ -129,7 +129,7 @@ def merge_statistics(stats_list):
   return list(name_map.values())
 
 
-def log_statistics_list(stats_list, logger_fn=tf.logging.info):
+def log_statistics_list(stats_list, logger_fn=tf.compat.v1.logging.info):
   """Calls the given logger function on each `Statistic` in the list.
 
   Args:


cghawthorne:
@adarob do you think we should just start switching to abseil logging?

The official [recommendation](https://github.com/tensorflow/community/blob/master/rfcs/20180827-api-names.md#deprecated-namespaces) seems to be to switch to the built-in Python logging, but I'm guess abseil will do a better job of more or less preserving API compatibility with the old tf.logging.

magenta/pipelines/statistics.py
```diff
@@ -129,7 +129,7 @@ def merge_statistics(stats_list):
   return list(name_map.values())
 
 
-def log_statistics_list(stats_list, logger_fn=tf.logging.info):
+def log_statistics_list(stats_list, logger_fn=tf.compat.v1.logging.info):
```
cghawthorne: Could you just change the import at the top of this file to: `import tensorflow.compat.v1 as tf`

cghawthorne:
Sorry for the delays in reviewing this, but I think we'll just go ahead and make the same change in #1614, so I'll close this PR. But thanks for your contribution, and we'd welcome further PRs from you!

PR #1597, opened 2019-10-01
apache beam now supports python 3

kousun12:
https://beam.apache.org/roadmap/python-sdk/#python-3-support

python 3.5 supported since 2.11.0 and python 3.6/3.7 supported since 2.14.0

this pr removes the python 2 conditionals for beam

### DIFF
--- a/magenta/models/score2perf/datagen_beam.py
+++ b/magenta/models/score2perf/datagen_beam.py
@@ -23,9 +23,7 @@ import functools
 import hashlib
 import logging
 import os
-import sys
 import random
-from absl import flags
 import apache_beam as beam
 from apache_beam import typehints
 from apache_beam.metrics import Metrics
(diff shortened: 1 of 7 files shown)


magenta/models/score2perf/datagen_beam.py
```diff
@@ -42,10 +41,11 @@
 # TODO(iansimon): this should probably be defined in the problem
 SCORE_BPM = 120.0
 
-FLAGS = flags.FLAGS
 flags.DEFINE_string(
     'pipeline_options', '',
     'Command line flags to use in constructing the Beam pipeline options.')
+FLAGS = flags.FLAGS
+FLAGS(sys.argv)
```
iansimon: Why is this needed?

kousun12: without it, usage later complains that we're trying to use the flag before it's been declared for some reason, and the specs fail. but actually it looks like the `tf.app.flags.FLAGS` wrapper rather than importing from `absl`. just made that change

PR #1600, opened 2019-10-05
Update README.md

falaktheoptimist:
Minor fixes in the README


magenta/models/onsets_frames_transcription/README.md
```diff
@@ -127,7 +127,7 @@ onsets_frames_transcription_infer \
   --examples_path="${TEST_EXAMPLES}" \
   --model_dir="${MODEL_DIR}" \
   --output_dir="${OUTPUT_DIR}" \
-  --hparams="use_cudnn=false" \
+  --hparams=use_cudnn=false \
```
cghawthorne: I actually like having the quotes here because it helps make it clear that "use_cudnn=false" is all one value for hparams=. Without it, the multiple ='s make it kind of hard to read.

falaktheoptimist: Agreed, makes sense. Done.

PR #1601, opened 2019-10-07
SavedModel export script for Music VAE

quantumpencil:
Script exports a saved model.pb with 3 signature defs, sample/encode/decode based on the interface in trained_model.py. Not sure if sample signature def is necessary as it is basically the same thing as decode, except z is not optional. 

TrainedModel's decode optionally returns decoder_results, but this is not a tensor so I'm confused as to what this is/if it is used. Will remove that comment after clarification

Upcoming PR will include a ServableVAEModel class which can be imported to query an instance of this savedmodel running in TF serving

### DIFF
--- a/magenta/models/music_vae/music_vae_export.py
+++ b/magenta/models/music_vae/music_vae_export.py
@@ -1,0 +1,219 @@
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+import argparse
+import json
+import tensorflow as tf
+from collections import namedtuple
+from tensorflow.python import saved_model
+from magenta.models.music_vae import configs
+
+
+# Function builds the VAE graph from a configuration object for exporting as a saved model
+def build_vae_graph(vae_config, batch_size, **sample_kwargs):
+    vae_graph = tf.Graph()
+    with vae_graph.as_default():
+        model = vae_config.model
+        model.build(
+            vae_config.hparams,
+            vae_config.data_converter.output_depth,
+            is_training=False)
+
+        # Define nodes for the model inputs
+        InputTensors = namedtuple(
+            "VAEInputTensors",
+            ["inputs", "controls", "inputs_length",
+             "c_input", "z_input", "temperature", "max_length"]
+        )
+        temperature = tf.placeholder(tf.float32, shape=())
+
+        if vae_config.hparams.z_size:
+            z_input = tf.placeholder(tf.float32, shape=[batch_size, vae_config.hparams.z_size])
+        else:
+            z_input = None
+
+        if vae_config.data_converter.control_depth > 0:
+            c_input = tf.placeholder(tf.float32, shape=[None, vae_config.data_converter.control_depth])
+        else:
+            c_input = None
+
+        inputs = tf.placeholder(
+            tf.float32,
+            shape=[batch_size, None, vae_config.data_converter.input_depth])
+        controls = tf.placeholder(
+            tf.float32,
+            shape=[batch_size, None, vae_config.data_converter.control_depth])
+        inputs_length = tf.placeholder(
+            tf.int32,
+            shape=[batch_size] + list(vae_config.data_converter.length_shape))
+
+        max_length = tf.placeholder(tf.int32, shape=())
+        input_tensors = InputTensors(inputs, controls, inputs_length, c_input, z_input, temperature, max_length)
+
+        # used in the trained_model decode functions, run in the stored session
+        OutputTensors = namedtuple(
+            "VAEOutputTensors",
+            ["outputs", "decoder_results", "mu", "sigma", "z"]
+        )
+        outputs, decoder_results = model.sample(
+            batch_size,
+            max_length=max_length,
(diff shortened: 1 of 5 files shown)


quantumpencil:
@googlebot I signed it!

magenta/models/music_vae/music_vae_export.py
```diff
@@ -0,0 +1,219 @@
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+import argparse
```
adarob: please use absl.flags

magenta/models/music_vae/music_vae_export.py
```diff
@@ -0,0 +1,219 @@
...
+        'sample': sample_signature_def,
+        'encode': encode_signature_def,
+        'decode': decode_signature_def
+    }
+
+    return signature_def_map
+
+
+# TODO create a named registry of signature def builders keyed by name for better type safety
+# Sample seems almost exactly like decode from the trained model code, possibly redundant
```
adarob: The difference is that sample feeds in a sample from a random normal distribution in place of the Z

magenta/models/music_vae/music_vae_export.py
```diff
@@ -0,0 +1,219 @@
...
+def build_sample_signature(input_tensors, output_tensors):
+    sample_inputs = {
+        'temperature': saved_model.utils.build_tensor_info(input_tensors.temperature),
+        'max_length': saved_model.utils.build_tensor_info(input_tensors.max_length)
+    }
+    if input_tensors.c_input is not None:
+        sample_inputs['c_input'] = saved_model.utils.build_tensor_info(input_tensors.c_input)
+    if input_tensors.z_input is not None:
```
adarob: as mentioned above, you should feed in a random normal sample for the z

magenta/models/music_vae/music_vae_export.py
```diff
@@ -0,0 +1,219 @@
...
+        method_name="sample"
+    )
+
+    return sample_signature_def
+
+
+def build_encode_signature(input_tensors, output_tensors):
+    # define the signature for the encoding operation
+    encode_inputs = {
+        'inputs': saved_model.utils.build_tensor_info(input_tensors.inputs),
```
adarob: use either ' or " throughout

magenta/models/music_vae/music_vae_export.py
```diff
@@ -0,0 +1,219 @@
...
+def build_decode_signature(input_tensors, output_tensors):
+    decode_inputs = {
+        'temperature': saved_model.utils.build_tensor_info(input_tensors.temperature),
+        'max_length': saved_model.utils.build_tensor_info(input_tensors.max_length),
+        'z_input': saved_model.utils.build_tensor_info(input_tensors.z_input)
+    }
+    if input_tensors.c_input:
+        decode_inputs["c_input"] = saved_model.utils.build_tensor_info(input_tensors.c_input)
+
+    # Question for magenta team about this, appears in trained model as a return value from decode,
```
adarob: These are really just used internally for the hierarchical decoder to be able to pass state around.

quantumpencil:
@adarob Thank you so much, I'll address these issues. I realized today that the CLA I signed actually needs to be a corporate one -- what process should I go through to void my individual CLA and sign the corporate one?

adarob:
You should be able to follow the instructions here: https://cla.developers.google.com/about/google-corporate

Let me know if you have issues.

quantumpencil:
Great, I just heard that the corporate CLA has been cleared/signed, so we're good to go.

magenta/models/music_vae/music_vae_export.py
quantumpencil: Done

magenta/models/music_vae/music_vae_export.py
quantumpencil: Cool -- I decided to just remove the signature as it's basically a matter of arguments, and the ServableModel class coming up will have handle passing an appropriately sampled random vector passed in to decode

magenta/models/music_vae/music_vae_export.py
quantumpencil: See above

magenta/models/music_vae/music_vae_export.py
quantumpencil: Great, removed

quantumpencil:
@adarob Sorry about the delay, had to push to cut an internal release. I've addressed your comments and made a couple small changes to config (a few new fields, and a method for serializing information needed in order to properly serve routines trained with the existing training infrastructure as part of the script)

The ServableModel/Servable VAE module I'll be PRing in uses the exported metadata when constructing the a new interface to the actual saved model running on tf-serving (simplified implementations of sample/interpolate in terms of the 'encode' and 'decode' signatures exported here, via PredictRequest proto over grpc) The new fields are strictly additive and default initialized so no existing code is affected.

Let me know if you have any other concerns.

PR #1608, opened 2019-10-23
fixed 'whether' typo

otacon94:
Fixed typo


otacon94:
@googlebot I signed it!

PR #1610, opened 2019-10-29
avoid CLOP by removing unwrap

drscotthawley:
phase unwrapping is not necessary to obtain instantaneous frequency, and/but can lead to catastrophic loss of precision (CLOP), i.e. errors that grow with time; especially noticeable for long audio clips.
In tests, this proposed change led to better waveform reconstruction accuracy and faster execution. 
Note that when reconstructing phase from dphase, cumsum should also be avoided as this too can lead to CLOP. 
cf. https://gist.github.com/drscotthawley/3c812fde932c649cd3cd41ea8a29803c

### DIFF
--- a/magenta/models/gansynth/lib/spectral_ops.py
+++ b/magenta/models/gansynth/lib/spectral_ops.py
@@ -196,10 +196,10 @@ def unwrap(p, discont=np.pi, axis=-1):
   return unwrapped
 
 
-def instantaneous_frequency(phase_angle, time_axis=-2, use_unwrap=False):
+def instantaneous_frequency(phase_angle, time_axis=-2, use_unwrap=True):
   """Transform a fft tensor from phase angle to instantaneous frequency.
 
-  Take the finite difference of the phase. Pad with initial phase to
+  (Unwrap and) Take the finite difference of the phase. Pad with initial phase to
   keep the tensor the same size.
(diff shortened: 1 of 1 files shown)


drscotthawley:
@googlebot I signed it!

jesseengel:
Cool, thanks for the contribution! This all seems fine, however it doesn't seem to be a limiting factor on audio quality for this model. Can you please set `use_cumsum=True` by default, so that the code still reflects the work of the paper by default?

Thanks!

drscotthawley:
Hi Jesse, I'm so honored that you'd check this out!   I'm happy to do as you ask, and will set `use_unwrap=True` and resubmit.  

I think you wouldn't notice any difficulty for short audio clips and/or relatively high numerical precision, but for longer clips (longer than 1 second) and/or lower precision (e.g. FP16), you'll really notice the error accumulation affect the audio.   In my tests with 3-second-long audio and FP32, unwrapping yielded a maximum waveform reconstruction error of around 1e-3, whereas with this 'different' method it was only 4e-6, and the difference was audible.  Plus it ran 20-40% faster. 

I only tested my method using numpy though (in the gist linked to above), not the full TF/magenta suite.   I'd love to hear if the altered version indeed works better for you, in terms of accuracy and/or execution time

PR #1613, opened 2019-11-02
Context manager handling potential memory leak

falaktheoptimist:
BytesIO stream created using with statement would handle closing of the stream in a cleaner fashion.

### DIFF
--- a/magenta/models/onsets_frames_transcription/data.py
+++ b/magenta/models/onsets_frames_transcription/data.py
@@ -181,9 +181,9 @@ def get_spectrogram_hash_op(spectrogram):
   def get_spectrogram_hash(spectrogram):
     # Compute a hash of the spectrogram, save it as an int64.
     # Uses adler because it's fast and will fit into an int (md5 is too large).
-    spectrogram_serialized = six.BytesIO()
-    np.save(spectrogram_serialized, spectrogram)
-    spectrogram_hash = np.int64(zlib.adler32(spectrogram_serialized.getvalue()))
+    with six.BytesIO() as spectrogram_serialized:
+      np.save(spectrogram_serialized, spectrogram)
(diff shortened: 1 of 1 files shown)


magenta/models/onsets_frames_transcription/data.py
```diff
@@ -181,9 +181,9 @@ def get_spectrogram_hash_op(spectrogram):
   def get_spectrogram_hash(spectrogram):
     # Compute a hash of the spectrogram, save it as an int64.
     # Uses adler because it's fast and will fit into an int (md5 is too large).
-    spectrogram_serialized = six.BytesIO()
-    np.save(spectrogram_serialized, spectrogram)
-    spectrogram_hash = np.int64(zlib.adler32(spectrogram_serialized.getvalue()))
+    with six.BytesIO() as spectrogram_serialized:
```
cghawthorne: It turns out this breaks py2 compatibility (which we're still supporting for at least a couple more months) because StringIO doesn't work with a context manager. Do you have suggestions for how to fix that? Also, do we know that the current code actually causes a memory leak? Wouldn't python ref counting take care of this?

falaktheoptimist: Understood. In that case, spectrogram_serialized.close() would work for both cases. As per [the docs](https://docs.python.org/3/library/io.html#io.BytesIO) 

> The buffer is discarded when the close() method is called. 

I wasn't sure about garbage collector/ reference counting working here because of it keeping things in the in memory buffer.

cghawthorne: ok yeah, switching to adding an explicit .close() call should be fine

falaktheoptimist: Cool. Done.

Issue #1574, opened 2019-08-03
Deprecation warnings


all0sound:
Just installed Magenta on Windows and tried the performance_rnn model with the Jupyter notebook (I had it running fine on a Mac about a year ago). Although it works, a good amount of deprecation warnings are displayed. Please let me know if there is something done or there is an update I did not know about. Thank you!

C:\Users\test\Anaconda3\envs\magenta\lib\site-packages\tensorflow\python\framework\dtypes.py:516: FutureWarning: Passing (type, 1) or '1type' as a synonym of type is deprecated; in a future version of numpy, it will be understood as (type, (1,)) / '(1,)type'.
  _np_qint8 = np.dtype([("qint8", np.int8, 1)])
C:\Users\test\Anaconda3\envs\magenta\lib\site-packages\tensorflow\python\framework\dtypes.py:517: FutureWarning: Passing (type, 1) or '1type' as a synonym of type is deprecated; in a future version of numpy, it will be understood as (type, (1,)) / '(1,)type'.
  _np_quint8 = np.dtype([("quint8", np.uint8, 1)])
C:\Users\test\Anaconda3\envs\magenta\lib\site-packages\tensorflow\python\framework\dtypes.py:518: FutureWarning: Passing (type, 1) or '1type' as a synonym of type is deprecated; in a future version of numpy, it will be understood as (type, (1,)) / '(1,)type'.
  _np_qint16 = np.dtype([("qint16", np.int16, 1)])
C:\Users\test\Anaconda3\envs\magenta\lib\site-packages\tensorflow\python\framework\dtypes.py:519: FutureWarning: Passing (type, 1) or '1type' as a synonym of type is deprecated; in a future version of numpy, it will be understood as (type, (1,)) / '(1,)type'.
  _np_quint16 = np.dtype([("quint16", np.uint16, 1)])
C:\Users\test\Anaconda3\envs\magenta\lib\site-packages\tensorflow\python\framework\dtypes.py:520: FutureWarning: Passing (type, 1) or '1type' as a synonym of type is deprecated; in a future version of numpy, it will be understood as (type, (1,)) / '(1,)type'.
  _np_qint32 = np.dtype([("qint32", np.int32, 1)])
[...]

PR #1614, opened 2019-11-02
Resolving deprecation warnings

falaktheoptimist:
Should handle couple of deprecation warnings pertaining to magenta from #1574

### DIFF
--- a/magenta/pipelines/statistics.py
+++ b/magenta/pipelines/statistics.py
@@ -129,7 +129,7 @@ def merge_statistics(stats_list):
   return list(name_map.values())
 
 
-def log_statistics_list(stats_list, logger_fn=tf.logging.info):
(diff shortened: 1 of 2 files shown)


magenta/pipelines/statistics.py
```diff
@@ -129,7 +129,7 @@ def merge_statistics(stats_list):
   return list(name_map.values())
 
 
-def log_statistics_list(stats_list, logger_fn=tf.logging.info):
+def log_statistics_list(stats_list, logger_fn=tf.compat.v1.logging.info):
```
cghawthorne: can you just change the import at the top of this file to `import tensorflow.compat.v1 as tf`?

falaktheoptimist: Yes, sure. Done.

PR #1621, opened 2019-11-12
Translate in Korean

bill0714:
for Korean developer


cghawthorne:
Thanks for your contribution! A few requests:

- Let's name the new file `README.ko.md` to be consistent with ISO 639-1 language codes and the file naming convention proposed by https://github.com/tiimgreen/github-cheat-sheet (which is as close as I could find to a standard for markdown files).
- Please add some text at the top of the Korean file explaining that it may be out of date because it is not officially maintained and that any questions about the translation should go to you.

PR #1623, opened 2019-11-13
Fixing keep amounts which were <0.5

falaktheoptimist:


cghawthorne:
I think we actually did train with these values (maybe not intentionally!), so I'd like to keep them as is. Have you tried training with these values vs. the changed ones?

falaktheoptimist:
Yes, we didn't get any onsets for a long time with the current values. Even when we tried to overfit by using a very small chunk of data (just 20 odd pieces). Fixing the dropout values fixed this problem. 
Maybe the API in TF changed from the version you used. But keep_prob of 25% will drastically slow down training I think.

cghawthorne:
OK, thanks for verifying. I suspect the API changed, especially since at this point we're using wrappers of wrappers of compatibility layers. :-)

PR #1629, opened 2019-11-23
Fix imports in MIDI interface.

hdanak:
Running the magenta_midi.py examples were failing with various import-related errors.


hdanak:
@googlebot I signed it!

PR #1640, opened 2019-12-04
Fix import in read_bundle_file

roshie548:
Importing `magenta.protobuf` throws an error, changed to `magenta.music.protobuf`.


roshie548:
@googlebot I signed it!

PR #1655, opened 2020-01-03
Fixed typo ("Choralies" corrected to "Chorales")

juliedotwav:


cghawthorne:
Thanks for the fix! Will merge once CLA is signed.

PR #1657, opened 2020-01-08
polyphony_rnn_train parameter fixed in README.md

ekstrah:
polyphony_rnn_train example parameter is wrong
instead of num_training_set, it has to provide num_eval_example
if not, it will produce an error giving batch_size is not greater than 0


ekstrah:
@googlebot I signed it!

PR #1671, opened 2020-02-23
Update README.md

mmontag:
Fix link to Polyphony RNN


mmontag:
@googlebot I signed it!

PR #1688, opened 2020-03-29
[WIP] Onset short samples

drzraf:
- More comments, clarify some variables and hardcoded values.
- Don't renumber the first thread as long as the audio doesn't overflow it
- debugging. More output, more details, keep track of inference time spent (for debugging purpose)
- ability to pad input-tensor if no enough samples [WIP]


drzraf:
@googlebot I signed it!

PR #1694, opened 2020-04-04
Minor code fixes

varshaneya:
1. `allow_pickle=True` flag is added for loading `npz` in order to avoid `Object arrays cannot be loaded when allow_pickle=False` error.
2. List does not have `.items()` as an attribute.


varshaneya:
@googlebot  I signed it!

PR #1709, opened 2020-04-29
Fix minor indentation error

pybnen:
return statement was inside the for loop in 'build_bidirectional_lstm'
method.


pybnen:
@googlebot I signed it!

adarob:
Oh wow. This is only minor if we're not trying to do multiple layers :) Thanks for finding it!

@iansimon we should verify that none of our models are using stacked bidirectional layers.

iansimon:
> @iansimon we should verify that none of our models are using stacked bidirectional layers.

Some of the 16-bar models were using multiple bidirectional layers :(

Issue #1708, opened 2020-04-29
Unable to run train command in sketch rnn


nirajpandkar:
**System**:
- Python: Python 3.7.7 (default, Mar 26 2020, 15:48:22)  [GCC 7.3.0] :: Anaconda, Inc. on linux
- Ubuntu 16.04 LTS
- Magenta 1.3.1

**Problem**:
Using the command
```
sketch_rnn_train --data_dir=dataset --gpu=0 --log_root=log --hparams="data_set=['eiffel_tower.npz']"
```
I get the following error -
```

/home/niraj/miniconda3/envs/magenta/lib/python3.7/site-packages/librosa/util/decorators.py:9: NumbaDeprecationWarning: An import was requested from a module that has moved location.
Import requested from: 'numba.decorators', please update to use 'numba.core.decorators' or pin to Numba version 0.48.0. This alias will not be present in Numba version 0.50.0.
  from numba.decorators import jit as optional_jit
/home/niraj/miniconda3/envs/magenta/lib/python3.7/site-packages/librosa/util/decorators.py:9: NumbaDeprecationWarning: An import was requested from a module that has moved location.
Import of 'jit' requested from: 'numba.decorators', please update to use 'numba.core.decorators' or pin to Numba version 0.48.0. This alias will not be present in Numba version 0.50.0.
  from numba.decorators import jit as optional_jit
INFO:tensorflow:sketch-rnn
I0429 08:03:50.716002 139879426516736 sketch_rnn_train.py:430] sketch-rnn
INFO:tensorflow:Hyperparams:
I0429 08:03:50.716129 139879426516736 sketch_rnn_train.py:431] Hyperparams:
Traceback (most recent call last):
  File "/home/niraj/miniconda3/envs/magenta/bin/sketch_rnn_train", line 8, in <module>
    sys.exit(console_entry_point())
  File "/home/niraj/miniconda3/envs/magenta/lib/python3.7/site-packages/magenta/models/sketch_rnn/sketch_rnn_train.py", line 471, in console_entry_point
    tf.app.run(main)
  File "/home/niraj/miniconda3/envs/magenta/lib/python3.7/site-packages/tensorflow_core/python/platform/app.py", line 40, in run
    _run(main=main, argv=argv, flags_parser=_parse_flags_tolerate_undef)
  File "/home/niraj/miniconda3/envs/magenta/lib/python3.7/site-packages/absl/app.py", line 299, in run
    _run_main(main, args)
[...]

PR #1710, opened 2020-04-30
sketch_rnn: Fixed an attribute error due to "six" module in python 3 and fixed loading of .npz datasets.

nirajpandkar:
Referencing this issue - #1708 

As discussed with @hardmaru , fixed 2 things - 
1. Commented these [two lines](https://github.com/tensorflow/magenta/blob/master/magenta/models/sketch_rnn/sketch_rnn_train.py#L432) to get rid of the attribute error which was occurring because of "six" module. Mostly a python2/3 issue
2. Added `allow_pickle=True`  parameter to the `np.load()` function without which training wouldn't proceed.

Trained successfully after making these changes :)

### DIFF
--- a/magenta/models/sketch_rnn/sketch_rnn_train.py
+++ b/magenta/models/sketch_rnn/sketch_rnn_train.py
@@ -130,10 +130,7 @@ def load_dataset(data_dir, model_params, inference_mode=False):
       data = np.load(six.BytesIO(response.content), encoding='latin1')
     else:
       data_filepath = os.path.join(data_dir, dataset)
-      if six.PY3:
-        data = np.load(data_filepath, encoding='latin1', allow_pickle=True)
-      else:
-        data = np.load(data_filepath, allow_pickle=True)
+      data = np.load(data_filepath, encoding='latin1', allow_pickle=True)
     tf.logging.info('Loaded {}/{}/{} from {}'.format(
         len(data['train']), len(data['valid']), len(data['test']),
         dataset))
--- a/magenta/models/sketch_rnn/sketch_rnn_train.py
+++ b/magenta/models/sketch_rnn/sketch_rnn_train.py
@@ -131,9 +131,9 @@ def load_dataset(data_dir, model_params, inference_mode=False):
     else:
       data_filepath = os.path.join(data_dir, dataset)
       if six.PY3:
-        data = np.load(data_filepath, encoding='latin1')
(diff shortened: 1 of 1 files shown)


nirajpandkar:
@googlebot I signed it!

magenta/models/sketch_rnn/sketch_rnn_train.py
```diff
@@ -131,9 +131,9 @@ def load_dataset(data_dir, model_params, inference_mode=False):
     else:
       data_filepath = os.path.join(data_dir, dataset)
       if six.PY3:
```
cghawthorne, line 133: we now support only py3, so no need to have these two paths

magenta/models/sketch_rnn/sketch_rnn_train.py
```diff
@@ -429,8 +429,8 @@ def trainer(model_params):
 
   tf.logging.info('sketch-rnn')
   tf.logging.info('Hyperparams:')
-  for key, val in six.iteritems(list(model_params.values())):
-    tf.logging.info('%s = %s', key, str(val))
+  # for key, val in six.iteritems(list(model_params.values())):
```
cghawthorne, line 432: can you either fix these lines or remove them?

nirajpandkar: Sure thing! I'll remove and recommit.

magenta/models/sketch_rnn/sketch_rnn_train.py
nirajpandkar: Ok! So is there a notification/message required to be given instead or should I remove it completely?

Marking @hardmaru too just in case.

magenta/models/sketch_rnn/sketch_rnn_train.py
nirajpandkar: Done! :+1:

magenta/models/sketch_rnn/sketch_rnn_train.py
nirajpandkar: Done! :+1:

nirajpandkar:
Is there anything else required from my side for the pull to be merged? @cghawthorne

cghawthorne:
Sorry for the delay, trying to get our internal merge processes to work correctly. It should be merged soon!

PR #1803, opened 2020-08-29
Modified to ignore vscode settings folder and all sub-files.

sungsikcho:
Previously, only files in the vscode folder were being ignored, but it would be more convenient to ignore the vscode folder as well. 

What do you think about this modification?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,5 +7,5 @@
 .cache/
 build/
 *.DS_Store
-.vscode/*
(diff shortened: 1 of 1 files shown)


sungsikcho:
I signed it!

2020년 8월 29일 (토) 오후 5:09, googlebot <notifications@github.com>님이 작성:

cghawthorne (approved):

PR #1822, opened 2020-09-23
Change read mode from default ("r") to "rb", fix "UnicodeDecodeError".

zhangxu999:
Change read mode from default ("r") to "rb", fix "UnicodeDecodeError".

PR #1823, opened 2020-09-23
fix "UnicodeDecodeError"

zhangxu999:
Change read mode from default ("r") to "rb",


cghawthorne (approved):

cghawthorne:
Going to close this one in favor of #1822 because this one has something weird about its commit history that's causing problems with our import process.

PR #1863, opened 2020-12-03
Minor tweaks

brianpow:
tweak message
change output filename to 'input filename prefix' + '.midi' instead of appending '.midi' to the input filename
add automatically audio conversion using ffmpeg if input file(s) is/are not wav

### DIFF
--- a/magenta/models/onsets_frames_transcription/onsets_frames_transcription_transcribe.py
+++ b/magenta/models/onsets_frames_transcription/onsets_frames_transcription_transcribe.py
@@ -115,7 +115,7 @@ def run(argv, config_map, data_fn):
         # the data processing functionality in data.py without having to
         # construct all the Example protos in memory ahead of time or create
         # a temporary tfrecord file.
-        tf.logging.info('Processing file...')
+        tf.logging.info('Processing file %s...' % (filename))
         sess.run(iterator.initializer,
                  {examples: [
                      create_example(filename, hparams.sample_rate,
--- a/magenta/models/onsets_frames_transcription/onsets_frames_transcription_transcribe.py
+++ b/magenta/models/onsets_frames_transcription/onsets_frames_transcription_transcribe.py
@@ -19,6 +19,7 @@ from __future__ import division
 from __future__ import print_function
 
 import os
+import subprocess
 
 from magenta.models.onsets_frames_transcription import audio_label_data_utils
 from magenta.models.onsets_frames_transcription import configs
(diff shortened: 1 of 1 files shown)


brianpow:
@googlebot I signed it!

magenta/models/onsets_frames_transcription/onsets_frames_transcription_transcribe.py
```diff
@@ -58,7 +59,13 @@
 
 def create_example(filename, sample_rate, load_audio_with_librosa):
   """Processes an audio file into an Example proto."""
-  wav_data = tf.gfile.Open(filename, 'rb').read()
+  fileparts = os.path.splitext(filename.lower())
+  if fileparts[1] != ".wav":
+    tf.logging.info('Converting %s into wav using ffmpeg...' % (filename) )
```
cghawthorne, line 64: rather than using subprocess here, what would you think of just calling out to `note_seq.audio_io.wav_data_to_samples_pydub`?

brianpow: The code is for converting non-wav file into wave. Tried `note_seq.audio_io.wav_data_to_samples_pydub` but it accepts wav data only.

cghawthorne: Ah, interesting. I'm surprised pydub couldn't correctly guess the format. If you add a format='mp3' parameter, does it work? If so, then yes, I think we should add that argument to `note_seq.audio_io.wav_data_to_samples_pydub`.

We should probably also change the name if we do that, but I can take care of that in a second pass because it affects some of our internal stuff too.

brianpow: Will you consider accepting subprocess instead since it is just a one-liner code? Refactoring `note_seq.audio_io.wav_data_to_samples_pydub` can be a long-term goal in the future.

cghawthorne: If possible, I'd really prefer the refactoring I suggested. Calling out to subprocess won't always work in our internal setup.

brianpow: pydub is also using `subprocess` to call `ffmpeg`, does it work in your internal setup?
btw, I believe this one-line code won't increase the burden to refactor the code in the future, and it benefits users who get tired to convert the music to wave first. It also won't break anything if you continue to use wave file as input.

Issue #1919, opened 2021-05-24
onsets and frames problems when eval mode


nkjulia:
hi i try to use onsets_frames_transcription_train.py and eval mode to output metrics on test data , but i got this error , is there anyone can help ? thanks~

INFO:tensorflow:evaluation_loop marked as finished
I0524 11:21:34.920045 140096573208320 error_handling.py:115] evaluation_loop marked as finished
WARNING:tensorflow:Reraising captured error
W0524 11:21:34.920181 140096573208320 error_handling.py:149] Reraising captured error
Traceback (most recent call last):
  File "magenta/models/onsets_frames_transcription/onsets_frames_transcription_train_maestro.py", line 133, in <module>
    console_entry_point()
  File "magenta/models/onsets_frames_transcription/onsets_frames_transcription_train_maestro.py", line 129, in console_entry_point
    tf.app.run(main)
  File "/DATA/disk1/anaconda3/envs/magenta/lib/python3.7/site-packages/tensorflow/python/platform/app.py", line 40, in run
    _run(main=main, argv=argv, flags_parser=_parse_flags_tolerate_undef)
  File "/DATA/disk1/anaconda3/envs/magenta/lib/python3.7/site-packages/absl/app.py", line 303, in run
    _run_main(main, args)
  File "/DATA/disk1/anaconda3/envs/magenta/lib/python3.7/site-packages/absl/app.py", line 251, in _run_main
    sys.exit(main(argv))
  File "magenta/models/onsets_frames_transcription/onsets_frames_transcription_train_maestro.py", line 124, in main
    additional_trial_info=additional_trial_info)
  File "magenta/models/onsets_frames_transcription/onsets_frames_transcription_train_maestro.py", line 113, in run
    num_steps=FLAGS.eval_num_steps)
  File "/DATA/disk1/juliayang/magenta-master/magenta/models/onsets_frames_transcription/train_util.py", line 253, in evaluate
    checkpoint_path=checkpoint_path, name=name)
  File "/DATA/disk1/anaconda3/envs/magenta/lib/python3.7/site-packages/tensorflow_estimator/python/estimator/tpu/tpu_estimator.py", line 3151, in evaluate
    rendezvous.raise_errors()
  File "/DATA/disk1/anaconda3/envs/magenta/lib/python3.7/site-packages/tensorflow_estimator/python/estimator/tpu/error_handling.py", line 150, in raise_errors
    six.reraise(typ, value, traceback)
[...]

PR #1934, opened 2021-06-17
fix errors when using eval mode

nkjulia:
fix errors when evaluate using script onsets_frames_transcription_train.py

related issue

#1919

### DIFF
--- a/magenta/models/onsets_frames_transcription/model.py
+++ b/magenta/models/onsets_frames_transcription/model.py
@@ -403,7 +403,7 @@ def model_fn(features, labels, mode, params, config):
       loss_label = 'losses/' + label
       tf.summary.scalar(loss_label, tf.reduce_mean(loss_collection))
     if mode == tf.estimator.ModeKeys.TRAIN:
-        train_op = slim.optimize_loss(
+      train_op = slim.optimize_loss(
         name='training',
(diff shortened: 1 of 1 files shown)


magenta/models/onsets_frames_transcription/model.py
```diff
@@ -133,7 +133,7 @@ def model_fn(features, labels, mode, params, config):
   length = features.length
   spec = features.spec
 
-  is_training = mode == tf.estimator.ModeKeys.TRAIN
+  is_training = (mode == tf.estimator.ModeKeys.TRAIN or mode == tf.estimator.ModeKeys.EVAL)
```
cghawthorne, line 136: `is_training` should be true only for the TRAIN mode. It affects things like whether dropout is enabled.

cghawthorne (changes requested):

magenta/models/onsets_frames_transcription/model.py
nkjulia: but with evaluate mode, parameters are fixed.

cghawthorne: Right, but dropout affects activations, so it will still cause problems if it's turned on during evaluation.

PR #1953, opened 2021-09-29
Fix scipy version error

junnei:
Fixed error that scipy.mics was deprecated.
Replaced scipy.misc to spicy.special


junnei:
- [ ]

PR #1970, opened 2022-02-16
Rl tuner compatibility  fix when usinf TF2

MrRobot2211:
Hi, there I have been playing around with RLTuner as part of a mentorship I am doing with Jordi Orbay. I added some fixes to make  `rl_tuner_train` run directly as advertised in the README.

- Added code to rename modules in the downloadable reward_rnn checkpoint
-  Added code to solve an issue with plot evaluation.

There are also changes to be done in the example notebook file, but maybe it is cleaner to do it in another PR. What do you think?


MrRobot2211:
@jorbay

PR #2004, opened 2022-07-01
Updated setup file

prasadjivane:
Modified setup.py


cghawthorne:
What's the goal of this change? Also, we typically use single quotes in our python files.

PR #2016, opened 2022-09-02
[sketch-rnn] - write hparams `model_config.json` as a dict not a list

silky:
this fixes what is evidently a long-standing bug in the way hparams were written by this model.

the fix is to use the `to_json` method provided by the `HParams` class itself. 

without this, the hparams are written as a list of the keys; not as a dictionary of key/values ... 😭


silky:
closing; i don't want this on my list on open PRs anymore.

